PR #47, opened 2015-04-01
File type icons

aaronshekey:
File-type icons were a bit too thick where the page folds. I've changed each to have a more consistent stroke weight on that top right corner.

file-type


cameronmcefee:
Looks good to me, @aaronshekey. Go ahead an merge and tag it when you're ready.

cameronmcefee:
Or leave it to me if you prefer.

PR #49, opened 2015-04-27
Correct the license ids used in bower.json

rtyley:
The bower.json spec [says that the license field](https://github.com/bower/bower.json-spec#license) should use [SPDX license identifiers](https://github.com/bower/bower.json-spec), so it turns out the correct license ids for Octicons are [`OFL-1.1`](https://spdx.org/licenses/OFL-1.1.html) and [`MIT`](https://spdx.org/licenses/MIT.html).

The current license text isn't recognised by automated tools like the on-demand [Bower WebJar](http://www.jamesward.com/2015/03/23/scaling-the-webjars-project-with-on-demand-bower-packages) generator, which blocks the distribution: https://github.com/webjars/webjars/issues/984

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,10 @@
 {
   "name": "octicons",
   "description": "GitHub's icon font",
-  "license": "SIL OFL 1.1, MIT",
+  "license": [
+    "OFL-1.1",
+    "MIT"
+  ],
   "homepage": "https://octicons.github.com",
   "authors": [
     "GitHub <support@github.com>"


cameronmcefee:
Thanks @rtyley!

@talniv, Can you confirm that this change is ok on the legal side?

rtyley:
The delta here is effectively `SIL OFL 1.1, MIT` --> `OFL-1.1, MIT`, and 'OFL' is an accepted abbreviation of [SIL Open Font License](http://en.wikipedia.org/wiki/SIL_Open_Font_License), so hopefully it has minimal legal impact?

cameronmcefee:
Sorry for the delay on this one. The discussion got caught up in some travel. Thanks @rtyley

rtyley:
Thanks @cameronmcefee! Could you cut a release off this so I can take advantage of the updated config?

cameronmcefee:
:+1: done

Issue #53, opened 2015-05-25
Add path (and version) variables to SCSS file


wjdp:
The LESS file has a variable `$octicons-font-path`, I'm using octicons in a project using Sass. It'd be nice to be able to set the font path in the SCSS file.

I've noted this repo is generated from somewhere else, but I'll still do a PR to show what's needed.

cameronmcefee:
Resolved in #54

PR #54, opened 2015-05-25
Add path and version variables to SCSS file

wjdp:
Regarding issue #53 these are changes required to add a path variable that can be overridden by the user.

I've noted this repo is generated from somewhere else, not sure how to proceed.


cameronmcefee:
:+1: thanks @wjdp. You're right, it's generated elsewhere (need to change that), but I merge these and then update the generator to spit out identical content so it doesn't create new changes in the future.

wjdp:
Ah sweet, no worries

PR #71, opened 2015-10-08
Document how to install via Homebrew

benbalter:
Per https://github.com/github/octicons/issues/70#issuecomment-146342298, documenting how to install Octicons via Homebrew. @cameronmcefee, @mikemcquaid anything else you'd like to see here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,15 @@ Octicons includes a stylesheet specifically for [Rails 4/Sprockets][sprockets].
     content_tag :span, '', :class => "octicon octicon-#{code.to_s.dasherize}"
   end
   ```
+  
+## Installing locally
+
+The easiest way to install Octicons locally is with Homebrew. Simply run the following commands:
+
+```
+brew tap "caskroom/fonts"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,6 +77,15 @@ Octicons includes a stylesheet specifically for [Rails 4/Sprockets][sprockets].
     content_tag :span, '', :class => "octicon octicon-#{code.to_s.dasherize}"
   end
   ```
+  
+## Installing locally
+
+The easiest way to install Octicons locally is with Homebrew. Simply run the following commands:
```
cameronmcefee: > The easiest way to install Octicons locally is with Homebrew.

I hesitate to agree with this, in that it requires you know what Homebrew is and that you have it installed. It might be better to say something like:

*It's easy to install octicons locally if you have [Homebrew](http://brew.sh/) installed.*

mikemcquaid: I agree.

mikemcquaid:
Instructions look correct.

benbalter:
Updated in 4427036.

PR #76, opened 2015-11-06
3.2.0

aaronshekey:
This adds the following Octicons:

`octicon-bold`
`octicon-text-size`
`octicon-italic`
`octicon-tasklist`

It also normalizes some styling in:

`octicon-list-unordered`
`octicon-list-ordered`
`octicon-quote`
`octicon-mention`
`octicon-bookmark`
`octicon-threebars`

Removes 

`octicon-screen-normal`
`octicon-screen-full`


vernmic:
why remove these? actually two very useful; use for my maximize and restore down icons on widgets for a dashboard I am building. 

octicon-screen-normal 
 octicon-screen-full 

just curious

aaronshekey:
Octicons is specifically designed for GitHub products. We don't have a need for those specific icons on GitHub, so we killed 'em. At GitHub's scale, it doesn't make sense to include icons that aren't actively used in our products. Our primary reason for releasing them was to make it easier for the people already ripping them out of our sites to use them in their own GitHub integrations. Sorry you were planning on using those 😕

PR #86, opened 2016-02-12
Bump to 3.5.0

aaronshekey:
### 3.5.0 (February 12, 2016)

Adds:

- `unverified`

Refines:

- `verified`


styfle:
Hi @aaronshekey,
It might be helpful to show the context in which GitHub uses the new icons that are added. I know there is the full featured https://octicons.github.com/ site but there have been times where you dropped icons in the past so it would be good to know where GitHub is using the new icons so other projects can use these icons in the same context. :ok_hand:

Issue #88, opened 2016-02-13
Create Octicons as <symbol>s


shgysk8zer0:
For use with `<use xlink:href...>`, create a single file (such as `octicons.svg`) containing all SVGs in the `svg/` directory as `<symbol>`s, having the filename as the ID and the `viewBox` set from height and width.

jonrohan:
fixed in https://github.com/primer/octicons/pull/89

Issue #87, opened 2016-02-13
Use viewBox


shgysk8zer0:
I'm not sure of any reasoning behind using `width` and `height` instead of `viewBox`. Whatever the reason, would it be possible to use `viewBox` either in addition to, or in place of, the dimensions?

jonrohan:
fixed in https://github.com/primer/octicons/pull/89

PR #89, opened 2016-02-14
Creating a grunt build system for octicons v4.0.0

jonrohan:
This is a WIP to create a grunt build system for octicons that will out put both font files and svg files.

The **tl;dr** of what I'm trying to do is:

* SVG icon files will be placed in `src/svg`
* The developer will run `grunt`
* All assets are compiled to `dist`

-
@aaronshekey @github/css


aaronshekey:
Yes! This is super awesome and addresses #88 and #87.

jonrohan:
@aaronshekey I have this working pretty well, but there's some weird hinting going on 
image need to fine tune it.

Plus here's a few more things I want to address

- [ ] Stuff labeled v4 https://github.com/github/octicons/milestones/v4.0.0
- [ ] Update site to show svg and font
- [ ] A little bit better organization of dist folder with styles and resources properly organized

aaronshekey:
Yeah, that is rendering strangely. Some of the unicode symbols have changed between this and prior versions, right? That _may_ be an issue, though I can't be sure how everyone's using Octicons. I'm not sure how high of a priority it is though.

mdo:
I haven't pulled down the code to see the full diff yet, but I'm seeing hella `_posts` files. How performant is that looking vs something like a data file?

jonrohan:
> but I'm seeing hella _posts files. How performant is that looking vs something like a data file?

I pulled that directly from https://github.com/github/octicons.github.com in the interest of combining all the 3 repos for octicons. I would be into updating it to use a data map though. :+1:

jonrohan:
> Some of the unicode symbols have changed between this and prior versions, right? That may be an issue

The beauty of a major update https://github.com/github/octicons#versions we're telling people it's breaking changes. `4.0.0`

aaronshekey:
> We're telling people it's breaking changes.

With this build system, any additional symbol will be a breaking change, right?

mdo:
> major update [...] breaking changes

🙌 🙌 🙌

jonrohan:
> With this build system, any additional symbol will be a breaking change, right?

I guess so, but we can create a character code map to keep them similar. https://github.com/sapegin/grunt-webfont#codepoints

hectoruch:
Why on the official page are that text? "Copypasta"

Thanks!

sophshep:
I think this is similar to what [Grunticon](https://github.com/filamentgroup/grunticon) does, thought they don't create a font. (Which is fine for our use internally, imo)

I just got that up and running in [a PR on the new help site](https://ghe.io/sophshep/new-help-templates/pull/5) and so far it is working great.

aaronshekey:
@sophshep I think we may indeed want to keep the font around for local Keynote presentations and the like. I've added the pre-established codepoints, so we should be covered :v:

jonrohan:
I'm going to merge this in, leaving the version at 3.5.0, not publishing anything new to npm. So I can do some testing in the github/github pipeline

PR #91, opened 2016-04-20
Version 4 authoring

### LINKED ISSUE #56: Compression octicons.css to octicons.min.css
Add compression octicons.css to octicons.min.css in the build from github.
Thanks.

aaronshekey:
This PR is a work in progress. It'll be merged into #89 and makes a few changes on the authoring side of Octicons.

Internally, we currently have a web app that accepts SVG and builds a font based on a json file. What I'd rather see is the following workflow:

1. An editable master Octicons Illustrator / Sketch file that exports raw SVG into the `src` directory
2. Grunt then processes those SVGs using a minifier to strip them of all the authoring tools nonsense, handle rounding, etc.
3. We build the fonts based on those processed SVGs

This PR is a step toward that. I did a small test with a couple Illustrator-generated SVGs and they seem to be working correctly.

Remaining tasks:

- [ ] Kill the batman
- [x] Build a master sketch document with all the octicons
- [x] We should offer minified CSS to close #56
- [ ] Offer a jekyll plugin to remove dependency on fonts
- [x] Update the `README` for accuracy
- [x] ~~We need to make sure we have an `octicons-local.ttf` of the same specs as the current library for our internal Keynote documents.~~ This is now possible with the latest version of `grunt-webfont`, but I'm not sure if it's completely necessary here.

--
cc: @jonrohan


jonrohan:
> Kill the batman

If this is so simple, why haven't you done it already?

aaronshekey:
@jonrohan How should we handle the jeykll plugin side of things? I think everything else in this PR is where it needs to be.

jonrohan:
> How should we handle the jeykll plugin side of things?

Let's get the building and output setup, then what I'm thinking is we can then turn this repo also into a gem that will include octicons_helper as well as the jekyll plugin in a new pr.

jonrohan:
> as well as the jekyll plugin in a new pr.

also, I'll probably need to research how most jekyll plugins are distributed, because gem might not be the way

aaronshekey:
Hi

Issue #98, opened 2016-06-16
Problem integrating with express app


jacargentina:
## Solution with 3.5.0

 I could integrate with my express website, by simply npm install octicons, and then by using the variable `$octicons-font-path`, which in my case is simply 

   `$octicons-font-path: ''`

That way i achieved the stylesheet like this:

```css
@font-face {
    font-family: 'octicons';
    src: url("/octicons.eot?#iefix&v=22dab811a5aa6c7ae0f71074a3480b993f7ef79b") format("embedded-opentype"),url("/octicons.woff?v=22dab811a5aa6c7ae0f71074a3480b993f7ef79b") format("woff"),url("/octicons.ttf?v=22dab811a5aa6c7ae0f71074a3480b993f7ef79b") format("truetype"),url("/octicons.svg?v=22dab811a5aa6c7ae0f71074a3480b993f7ef79b#octicons") format("svg");
    font-weight: normal;
    font-style: normal
}
```

Please note the / at the start of the url for the .ttf font, etc

Also, i was serving the /octicons.ttf directly from

    node_modules/octicons/octicons/octicons.ttf

By doing: 

    express.static('node_modules/octicons/octicons');

## Problem with 4.1.0

Now the `octicons-font-path` variable is not anymore. And that makes my life miserable, as now the css generated does not contains the / at the beginning. But its a simple relative to the current url path reference, like this:

```css
@font-face {
	font-family:"Octicons";
	src:url("octicons.eot?e8b041c217b3d809c380606d0320205a");
	src:url("octicons.eot?#iefix") format("embedded-opentype"),
		url("octicons.woff2?e8b041c217b3d809c380606d0320205a") format("woff2"),
		url("octicons.woff?e8b041c217b3d809c380606d0320205a") format("woff"),
		url("octicons.ttf?e8b041c217b3d809c380606d0320205a") format("truetype"),
		url("octicons.svg?e8b041c217b3d809c380606d0320205a#octicons") format("svg");
	font-weight:normal;
	font-style:normal;
}
```
Note the missing / at the beginning.

So now, my express.static is useless, and i cant serve the .ttf font as i was doing, cause the browser tries to fetch the .ttf from the same base url as my .css stylesheet, which in my case is `/static/css/all.css`

So, it tries to load octicons.ttf from:

    `GET /static/css/octicons.ttf`

And the express.static i've already pointed out then searchs for:

   `node_modules/octicons/octicons/static/css/octicons.ttf`

[...]

PR #99, opened 2016-06-16
Adding the path variable to the scss file

jonrohan:
fixes https://github.com/primer/octicons/issues/98

I'm adding a `$octicons-font-path: "./" !default;` as the default for the `_octicons.scss` font stylesheet. This let's the user set it themselves by overriding it in their import sheet. For example:

```scss
$octicons-font-path: "./path/to/font";
@import "octicons/build/font/octicons";
```

-
@jacargentina Does this work for you?
cc @aaronshekey


jacargentina:
Trying it right now.

jacargentina:
:+1: All works again with 4.1.0. Thanks @jonrohan!

Issue #97, opened 2016-06-13
Profile and orgs are very masculine


aaronshekey:
Let's reconsider how these icons are drawn. They're super masculine, and I'm not sure we need the entire body.

![screen shot 2016-06-13 at 12 50 32 pm](https://cloud.githubusercontent.com/assets/1369864/16017439/808d0a1e-3165-11e6-8661-14d3198aea4f.png)
![screen shot 2016-06-13 at 12 50 43 pm](https://cloud.githubusercontent.com/assets/1369864/16017440/809f594e-3165-11e6-9435-3a793d231920.png)

--
@jglovier

jglovier:
Just for reference: https://thenounproject.com/term/gender-neutral/147092/

jglovier:
More reference:

Twitter app:

![image](https://cloud.githubusercontent.com/assets/1319791/16022407/80c8eba4-316c-11e6-8378-df339d7ece4c.png)

Instagram app:

![image](https://cloud.githubusercontent.com/assets/1319791/16022422/91348516-316c-11e6-8d66-e745b531b1fd.png)

aaronshekey:
I think there's a way we can remove hair and the body without having shoulders that are too broad. When noodling around on some UI, I explored something like this:

![screen shot 2016-06-13 at 3 54 39 pm](https://cloud.githubusercontent.com/assets/1369864/16022781/2bc2f75e-317f-11e6-9ac9-dde5188c9602.png)

There are lots of established patterns for this, and we just need to define which one we like best aesthetically without anybody feelin' excluded.

jglovier:
I like the idea of a head as a full circle because it avoids any hinting at actual facial characteristics, and with a half circle for the bust it wouldn't feel like broad shoulders.

Something like this:

<img width="345" alt="screenshot 2016-06-13 17 47 47" src="https://cloud.githubusercontent.com/assets/1319791/16027905/024d2e84-318f-11e6-9961-b77bbe0de7fb.png">

PR #103, opened 2016-06-26
Slide the italic shape up a bit in its container

aaronshekey:
This PR addresses a tiny alignment issue with the `italic` shape. It updates the master Octicons file, and rebuilds with that new shape.

I'm good with this living in `HEAD` on `master` for a bit. No need to cut a new release just yet, unless we feel like it!

#### Before
screen shot 2016-06-26 at 12 23 25 pm

#### After
screen shot 2016-06-26 at 12 22 42 pm

--
@jonrohan


jonrohan:
Cool 👍 I don't have any immediate commits I'll be making to the repo, so if you wanted to wait to release and tackles some of the other issues #97 then I'd be game.

PR #106, opened 2016-06-30
Removing the inline sourcemap from the build files

### LINKED ISSUE #105: Is it intentional that octicons.min.css is larger than octicons.css?
The .min.css file is minified, and contains no insignificant whitespace or comments.

However, it does contain an inlined (as base64 data) sourceMappingURL comment, which adds significantly to its size.

I can't be sure if this is on purpose or an unintended bug.

https://github.com/primer/octicons/blob/master/build/font/octicons.css - 8.93 KB
https://github.com/primer/octicons/blob/master/build/font/octicons.min.css - 26.7 KB

jonrohan:
fixes #105 

The grunt build cssnano process was adding inline sourcemaps to the output. Removing.


shurcooL:
LGTM.

PR #109, opened 2016-07-10
New `organization` and `person`

aaronshekey:
This PR hopes to address #97. I like them aesthetically. They are less gendered. They aren't so outlined and blocky.

screen shot 2016-07-10 at 6 00 22 pm

screen shot 2016-07-10 at 6 04 42 pm

--
@jonrohan @jglovier


jglovier:
Looks good. Curious about how/when we decide to use outline vs solid?

Also, mind posting a giant one for reference?

aaronshekey:
> Curious about how/when we decide to use outline vs solid?

It's pretty arbitrary at this point, but most real-world objects end up looking solid in our icon set. Sometimes its just down to an aesthetic choice :smile:

screen shot 2016-07-11 at 10 11 05 am
screen shot 2016-07-11 at 10 11 38 am

jglovier:
I def think this is better than the existing, but the more defined form feels a little out of place in our current set. What about using just a circle for the head, maybe detached from the shoulders? More geomtric like?

aaronshekey:
@jglovier I tried it and really wasn't feeling it in earlier commits. It was feeling very foreign and still quite masculine. We don't really have anything that's so primarily geometric `¯\_(ツ)_/¯`

mdo:
Snap, didn't know we were revisiting these glyphs until they after they shipped! Overall I'm a big fan of moving away from our former robot looking people outlines. Building off what @jglovier said, the filled in people here feels a bit off when viewing my profile because nothing else there is filled.

Note the tabs at the top and the profile deets in the left sidebar. Feels much less cohesive now.

screen shot 2016-07-12 at 8 49 32 pm

aaronshekey:
I can't seem to team mention across orgs, unfortunately 😞. I think we can also fill in the location icon to break up the icons here. I can also explore an outlined version, but I don't think it'd be appropriate for the glyphs here.

In this particular view, I'm more concerned with the vertical alignment :wink:

PR #118, opened 2016-09-09
Stop checking the build directory into the repo

jonrohan:
This pull updates the `.gitignore` and `.npmignore` so that the `/build/` directory is not checked into the repository, but it's included in the package published to npm.

The advantages to doing this are

* We don't have two `/svg/` folders confusing users
* Making edits to `/build/` files often creates conflicts


arthurhammer:
Quick heads up that the readme needs to be updated for this. Just stumbled over this looking for `build` 😅: 

> All the files you need will be in the /build/ directory already...

jonrohan:
Thanks for the ping, I have an update to the readme ready in this PR https://github.com/primer/octicons/pull/120 but I'm going to update it to make sure that I clearly say the `/build/` directory is the minimized and compiled files and it's located in npm or can be built with the repo.

PR #122, opened 2016-09-13
Using multiple sized octicon vectors

jonrohan:
Currently our octicons api is a bit naive. We draw all our icons at on a `16x16` pixel grid. This works perfectly for cases where we render the icon at 16px. But when we're rendering the icons at larger sizes ie. 32, 64. The vectors are just scaled up versions of the 16px icon.

With this pull that all changes. The goal is to accomplish this without any changes to how the developer interacts with Octicons. Now, when the developer requests an icon at any size, the code uses the SVG source of the closest size.

For example, requesting an icon at `34px`, will return the path drawn at `32px` and scaled up.

```
> var octicons = require("./index.js")
> octicons.alert.toSVG({ height: 34 })
'<svg version="1.1" class="octicon octicon-alert" aria-hidden="true" height="34" width="34" viewBox="0 0 32 32"><path d="M17.3 23.86a1.44 1.44 0 0 1-1.44 1.44h-.72a1.44 1.44 0 0 1-1.44-1.44v-.72c0-.795.645-1.44 1.44-1.44h.72c.795 0 1.44.645 1.44 1.44v.72zM14 10.32c0-.73.537-1.32 1.2-1.32h.6c.663 0 1.2.59 1.2 1.32v8.36c0 .73-.537 1.32-1.2 1.32h-.6c-.663 0-1.2-.59-1.2-1.32v-8.36zm16.721 15.653L17.247 2.998c-.78-1.33-2.713-1.33-3.494 0L.279 25.973C-.508 27.315.465 29 2.027 29h26.946c1.562 0 2.535-1.685 1.748-3.027z"/></svg>'
```

There's no changes to the api for this to work, it just does™

The source files are located in `/lib/svg/[SIZE]/` The folder size denotes the size that will be rendered. What the code does is look for the size that is less than and closest to the requested size. So requesting `34` return `32` requesting `16` returns `16`. And if the number is too small or there aren't any other sizes, it defaults to returning `16`

@primer/design-systems @aaronshekey


jonrohan:
@aaronshekey if this is merged in, we'll need to merge the icon sketch file with the one in your branch. Currently the one in this branch is just from master, so it doesn't have the larger icons in it.

sophshep:
I love the idea of having custom icons depending on size, I can already think of a bazillion places that this will be useful. Some questions I have:

- How will marketing icons [like these](https://cloud.githubusercontent.com/assets/527589/18492947/dc4e5ccc-79c2-11e6-8583-deaa00238c6b.png)
 fit within this? I assume that our core octicons should be just for UI/product icons, but want to make sure we're all on the same page there. Should we have a separate set of icons for this style?
- I chatted with @mdo this morning about our approach for designing icons and he suggested we work with a ~professional iconographer~, which I think would be awesome. I have worked with [Brian Brasher](http://etherbrian.org/) in the past and he was great. What do you all think about bringing him (or someone else) in to help with these?

jonrohan:
> How will marketing icons like these fit within this? I assume that our core octicons should be just for UI/product icons, but want to make sure we're all on the same page there. Should we have a separate set of icons for this style?

Even though they are technically the same technology as the octicons, I would classify those as a more specific set of illustrations. Which we might not want to open source. Core octicons I think should remain UI/product icons. 

If we want or see a need, I can build tooling around those also. 👍 

> What do you all think about bringing him (or someone else) in to help with these?

💯 Would love to have a full design session concerning the art and direction of all our octicons. But this PR is less concerned with the look of multiple icons, than the tooling around having extra icons for larger sizes. 

I purposefully wrote this change to take into account that some (maybe all) icons won't have larger multiple sizes. So we can add extra overrides as we see fit. All we'll need to do is drop icons into a new folder that represents the optimum size the vector was designed for.

```
/lib/svg/
      ├── 16/
      ├── 32/
      ├── 48/
      └── 660/
```

The only folder we'll assume that has all the icons is the `/16/` folder, because that's the size the octicons were designed at.

In the interest of moving these changes forward and eventually a new release I can remove the few extra icons I added here, so we can create a plan for design updates. What do you think?

aaronshekey:
@sophshep I'm currently plugging along with having 3 total sizes. Small, Medium, and Marketing. I've got a branch called `five-authoring` where I'm creating artboards for each individual size. It'll be laborious, but I think once we get over the initial hump, it'll be easy to maintain.

screen shot 2016-09-13 at 8 40 50 pm

I'd _love_ to hire some help, and I have lots of ideas for who we may consider, but perhaps we take that conversation offline ☺️

Issue #130, opened 2016-10-07
Add a Projects icon


jasonmassey:
Could use the icon for Projects!

PR #132, opened 2016-10-10
Projects icons

aaronshekey:
Adding a couple icons that we used in the release of Projects. This is branched from the 4.4.0 tag, and could make a 4.5.0 release since we're adding new shapes.

@jonrohan


shurcooL:
Does this PR resolve #130? If so, it would be nice to add "Resolves #130." so that issue is closed and referenced when this is merged.

jonrohan:
closing in favor of https://github.com/primer/octicons/pull/137

Issue #138, opened 2016-10-29
Projects icon not compiled well


jonrohan:
Just discovered this bug, the projects icon when it was optimized got turned into one big square. 

<img width="804" alt="screen shot 2016-10-28 at 9 40 24 pm" src="https://cloud.githubusercontent.com/assets/54012/19826402/2747dfb6-9d57-11e6-8bee-4edcde6fba0e.png">

PR #139, opened 2016-10-29
Leaving fill-rule on minimized SVG elements

jonrohan:
fixes https://github.com/primer/octicons/issues/138

Removing the `fill-rule` attribute will make the projects icon become a block. Leaving it on properly renders it. The file size is slightly larger, but until we can do an audit and redesign our icons, we'll need to have this to make it work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 .DS_Store
 build
+*.log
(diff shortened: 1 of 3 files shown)


shurcooL:
> fixes #6

Did you mean #138? #6 is a merged PR and appears to be unrelated.

jonrohan:
@shurcooL yup, thanks for the 👀

PR #144, opened 2016-11-13
Update README sentence with toSVG

### LINKED ISSUE #143: Outdated sentence on README
> There will be a key for every icon, with `keywords` and `svg`.

`svg` is now `toSVG()`

1j01:
I replaced the specific mention of `keywords`, because I'm guessing it was one of only a few properties

Fixes #143


jonrohan:
Thanks again @1j01

PR #145, opened 2016-11-13
Minor README normalizations

1j01:
A/an, punctuation, capitalization of npm, (simple form of) relative directories, and consistent "Returns x."

Moved `--save` to the end of the command line so you can copy the install command without it if you want (even though you generally always want it)


jonrohan:
Thanks @1j01 ❤️

PR #148, opened 2016-11-17
Keynote Template

sophshep:
This adds a keynote template for our octicons, so that people can copy & paste vector icons in their presentations.

For now these are only in white, but I will soon add black and colors as needed after that.

cc/ @primer/design-systems


jonrohan:
Thanks for this @sophshep I think I'd like to move this to a separate repository specifically focused on keynote. I can do that for you, just wanted to let you know

jonrohan:
Added here https://github.com/primer/keynote-octicons

sophshep:
> I think I'd like to move this to a separate repository specifically focused on keynote.

@jonrohan 👍  Can we document this somewhere so that when hubbers are looking for them they know where to find them?

PR #163, opened 2017-06-02
Replace ES6 Object.assign method with polyfill

nickpeihl:
[Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) is an ES6 method and is not supported on older browsers such as Internet Explorer and the Android mobile browser. This PR adds a [commonly-used module](https://dependency.land/object-assign/*) for a polyfill to fall back on if the user is using an older browser. If the user has an evergreen browser the native Object.assign method is used.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,11 +1,12 @@
 var data = require('./build/data.json')
+var objectAssign = require('object-assign')
 
 Object.keys(data).forEach(function(key) {
 
   // Returns a string representation of html attributes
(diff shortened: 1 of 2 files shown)


jonrohan:
Hey thanks for the PR, sorry it took so long to merge in. We were debating on if we should only support node or browser use is fair. We decided it wouldn't hurt to include it. Going to be in the next release

Issue #159, opened 2017-03-07
Two Ellipse Icons - Duplicate


GladLock:
I noticed you have two ellipse icons of the exact same thing, I assume it's just because of a misspelling:

.octicon-ellipsis 
.octicon-ellipses

shurcooL:
They do look identical.

- https://octicons.github.com/icon/ellipsis/
- https://octicons.github.com/icon/ellipses/

GladLock:
Is one meant to be vertical and the other horizontal?

djch:
In case anyone's confused about the grammar: 'ellipsis' is singular and 'ellipses' is plural, so 'ellipsis' would be more accurate I imagine.

dkfiresky:
I think it's just an alias for `.octicon-ellipsis`

cmwinters:
Not sure what the origin of this duplicate is, but thanks for pointing it out!

jonrohan:
Yup, this was some kinda dupe mistake. Got a pull to delete one.

PR #164, opened 2017-06-12
Deleting ellipsis typo

jonrohan:
fixes https://github.com/primer/octicons/issues/159

This removes the `ellipsis` icon which is a typo of the `ellipses` icon. They're the same thing.

@primer/design-systems


sophshep:
Should probably also update the version here, since it has the potential to break in places where people are using the icon we're deleting.

shurcooL:
FWIW, I (a passerby) agree with @sophshep that it makes more sense to keep `ellipsis` ([singular version](https://www.google.com/search?q=define+ellipsis), a set of 3 dots) and remove `ellipses` (plural) instead. @djch also mentioned this in https://github.com/primer/octicons/issues/159#issuecomment-284924868.

shurcooL:
Yay, thanks for fixing this issue! 🎉

PR #169, opened 2017-08-04
Add kebab icons

jglovier:
This PR adds a kebab icon. Actually two - a horizontal version and a vertical version. We will likely standardize our use on the horizontal version, but it seems prudent to include a vertical version for instances outside our .com implementation (such as Atom, Desktop, or other non-standard .com instances).

image


jglovier:
Pushed alignment with pixel grid per discussion with @pmarsceill and @jonrohan in Slack.

image

shurcooL:
How does this icon compare to [ellipsis](https://octicons.github.com/icon/ellipsis/)?

If I understand this right, this is a variation of the hamburger icon typically used to expand/collapse menus, is that right?

jglovier:
>How does this icon compare to ellipsis?

@shurcooL ellipses is more for toggling inline content, and the kebab is more specific for showing a menu of additional options. Although we may end up modifying the use of the ellipses icon with this addition as well.

PR #176, opened 2017-09-22
Add icon server for SVG + PNG

shawnbot:
This adds an [express](https://expressjs.com/) server when you run `npm start` that will serve up octicons at different sizes at URLs in the form:

```
/{symbol}.(svg|png){?params}
```

Where `params` are passed along as `octicons[symbol].toSVG(params)` and become attributes of the generated `<svg>` element, e.g.

```
/smiley.svg?width=400&fill=red
```

By default, it runs on a random open port. You can change that by setting the `PORT` environment variable:

```sh
PORT=4001 npm start
```

which should make the above URL available at:

```
http://127.0.0.1:4001/smiley.svg?width=400&height=400&fill=red
```

~From here, it should be pretty trivial to add a PNG endpoint that rasterizes the SVG.~ [sharp](http://sharp.dimens.io/en/stable/) made it ridiculously easy to add PNG rasterizing! 🎉


jonrohan:
We moved this to another repo

PR #179, opened 2017-10-13
Monocticons: Octicons mono repository

jonrohan:
After trials and errors with the [Mono-repository in primer-css](https://github.com/primer/primer-css/pull/230) We have decided that doing the same thing with octicons would be beneficial to our workflow.

This PR updates the the [octicons repositories](https://github.com/primer?utf8=%E2%9C%93&q=octicons&type=&language=) to modules in one repo.

While working on this, @shawnbot and I accidentally published the next version of octicons 7.0.0. So the release notes are included in this PR.


.github/CONTRIBUTING.md
```diff
@@ -1,3 +0,0 @@
-The contents of */octicons* */svg* are generated by an automated process. Changes to these files may be accepted, but may also be overwritten.
-
-Octicons are a scalable set of icons created by GitHub. At this time, new icons will only be added when they are needed for GitHub products.
```
shurcooL: > At this time, new icons will only be added when they are needed for GitHub products.

Is this statement still accurate or not anymore? It was removed here, and I can't find it re-added anywhere else.

PR #189, opened 2017-12-11
Add report icon

katmeister:
Via the discussion in https://github.com/github/design-systems/issues/435, here's the final `report` octicon! Could use a final pass to make sure the Sketch file and pixels are 👌 

Thank you!

cc @dmleong for contributor-flagged content visibility 💥

PR #190, opened 2017-12-12
Use so-sketchy to version sketchtool

max:
- Sometimes `sketchtool` has bugs (I see you 48.1).
- Versioning software is nice.
- NPM packages are nice.

`so-sketchy` is an NPM module that vendors in `sketchtool` and adds semantic versioning.


max:
This PR should fix some issues that @katmeister was having in #189.

script/install-sketchtool
```diff
@@ -1,19 +0,0 @@
-#!/bin/bash
```
jonrohan: We should remove `script/install-sketchtool` from `script/bootstrap` also.

max: Whoops, my bad. Done!

PR #221, opened 2018-05-09
CHANGELOG.md: Fix typos in removed icon names.

shurcooL:


shurcooL:
No problem. I noticed that v7.3.0 was missing those 2 icons when upgrading from v6.0.1, so I wanted to check the CHANGELOG to confirm it was intentional. That’s when I spotted the typos.

PR #222, opened 2018-06-08
React Octicons!

shawnbot:
This is a rehash of #196 introducing a React component API that looks like this:

```jsx
import Octicon, {Zap} from '@github/react-octicons'

export default () => <Octicon><Zap /></Octicon>
// or
export default () => <Octicon icon={Zap} />
```

With credit to @hiendv for the inspiration in  https://github.com/primer/octicons/pull/196#issuecomment-377674252 :beers:

Note that there's one wild assumption in the generated per-icon components here, and it's that the SVG from the `octicons` package is valid XML, which is what allows us to use it in JSX without munging it. I've checked the generated icon component JS into git so that we can look at it without the module having to publish.

### TODO
- [x] Add new tests for React Octicons
- [x] Get tests passing
- [x] Update `README.md` with new API (see [rendered](https://github.com/primer/octicons/tree/reocticons/lib/octicons_react/README.md))
- [x] Deploy [test site](https://githubocticons-react-ruacygemxe.now.sh/)


hiendv:
Thanks for the mention. For me, tree-shaking is a must for any kind of collection libraries.

- These collections may expand forever. People don't usually use the whole set of octicons.
- People don't usually use a single icon library.

Yes, there are projects using all kind of icons. Unless we provide an optimal way to use octicons, these projects will end up with a mess.

Disclaim: I'm experimenting the tree-shaking version of Octicons and libraries for React and Vue.js.

emplums:
@shawnbot I'm into it! 👍

lib/octicons_react/__tests__/all.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import {Alert, Zap, getIconByName, iconsByName} from '../'
+import renderer from 'react-test-renderer'
+
+const render = el => renderer.create(el).toJSON()
+
+describe("import {getIconByName}", () => {
+  it('gets named icons', () => {
+    const Icon = getIconByName('zap')
+    expect(render(<Icon/>)).toEqual(render(<Zap/>))
```
shawnbot: This doesn't need to render. It should just `expect(getIconByName('zap')).toEqual(Zap)`.

lib/octicons_react/script/build.js
```diff
@@ -0,0 +1,74 @@
...
+    const file = join(DIR, key + '.js')
+    return fse.writeFile(file, code, 'utf8')
+      .then(() => ({key, name, file}))
+  })
+
+function writeIndex(icons) {
+  const count = icons.length
+  console.warn('wrote %d files to %s.', count, DIR)
+  const file = join(DIR, 'index.js')
+  const code = `/* THIS FILE IS GENERATED. DO NOT EDIT IT. */
```
shawnbot: I wonder if we should say where/how this is generated here?

lib/octicons_react/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "react": "^16.4.0",
+    "prop-types": "^15.6.1"
+  },
+  "devDependencies": {
+    "babel-preset-env": "^1.7.0",
+    "babel-preset-react": "^6.24.1",
+    "fs-extra": "^6.0.1",
+    "jest": "^23.2.0",
+    "next": "^6.0.3",
+    "octicons": "7.2.0",
```
shawnbot: Note: `octicons` is indeed the source of truth, but it's a dev dependency because we build the icon components from the package rather than using a runtime dependency. No innerHTML, no parsing, just pure JSX.

lib/octicons_react/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "icons",
+    "index.js",
+    "pages"
+  ],
+  "scripts": {
+    "start": "next",
+    "test": "jest",
+    "publish": "../../script/notify success",
+    "prepublishOnly": "../../script/notify pending",
+    "prepare": "script/build.js"
```
shawnbot: Running this on `prepare` is not right, because it will run on local installs and all of the dependencies that it needs to build are dev-only. I think this should be `prepublishOnly` instead, which will fail `npm publish` if it generates a diff.

lib/octicons_react/script/build.js
jonrohan: This is a good idea do you think you could add it to all the generated .js files that are checked in?

script/export
```diff
@@ -18,6 +18,16 @@ try {
...
 
+const {
+  FIGMA_DOMAIN = 'api.figma.com',
+  FIGMA_TOKEN
+} = process.env
+
+if (!FIGMA_TOKEN) {
+  console.error('You must set FIGMA_TOKEN to a Figma personal access token.')
+  process.exit(1)
+}
```
jonrohan: nice improvements

lib/octicons_react/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "@github/octicons-react",
+  "version": "0.1.0",
```
jonrohan: should this version match up with what's in `lerna.json`?

shawnbot: Yes, definitely.

lib/octicons_react/src/index.js
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  if (width && height) {
+    Object.assign(attrs, {width, height})
+  } else {
+    const dims = {width: widthHeight[0], height: widthHeight[1]}
+    const given = width ? 'width' : 'height'
+    const computed = given === 'width' ? 'height' : 'width'
+    attrs[given] = width || height || sizeMap[size] || size
+    attrs[computed] = attrs[given] * (dims[computed] / dims[given])
```
shawnbot: If both the width and height aren't provided, we compute the width or the height — depending on whether they provide `width`, `height`, or `size` (which can be either a number or a string that's mapped via `sizeMap`).

lib/octicons_react/src/index.js
```diff
@@ -0,0 +1,71 @@
...
+  verticalAlign: 'text-bottom'
+}
+
+Octicon.propTypes = {
+  ariaLabel: PropTypes.string,
+  children: PropTypes.element,
+  height: PropTypes.number,
+  icon: PropTypes.func,
+  size: PropTypes.oneOfType([PropTypes.number, PropTypes.oneOf(Object.keys(sizeMap))]),
+  verticalAlign: PropTypes.oneOf(['middle', 'text-bottom', 'text-top', 'top']),
```
shawnbot: Is there a good reason not to allow `bottom` here?

lib/octicons_react/src/index.js
shawnbot: Note: the parens on that last line above are a linting requirement. `*` and `/` have the same operator precedence, so I think the rule that wants one expression grouped exists to help us be more explicit about what it's doing.

lib/octicons_react/package.json
```diff
@@ -5,9 +5,11 @@
   "homepage": "https://octicons.github.com",
   "author": "GitHub, Inc.",
   "license": "MIT",
-  "main": "index.js",
+  "main": "build/index.js",
+  "module": "index.js",
```
shawnbot: This was wrong for me to include. I removed it in 02a007a.

lib/octicons_react/.babelrc
```diff
@@ -0,0 +1,8 @@
+{
+  "presets": ["env", "stage-0", "react"],
+  "env": {
+    "production": {
+      "presets": ["next/babel"]
```
shawnbot: This is used when we call `NODE_ENV=production next` in the `start` run-script.

lib/octicons_react/.eslintignore
```diff
@@ -0,0 +1 @@
+src/icons/*.js
```
shawnbot: Don't lint the generated files. 🙈

lib/octicons_react/.nvmrc
```diff
@@ -0,0 +1 @@
+8
```
shawnbot: This is for development. I've added the equivalent to `engines.node` in the `package.json` as well.

PR #223, opened 2018-06-29
Minor release 7.4.0

shawnbot:
Tracking Issue for next major release: 📦 `7.4.0`
Approximate release date: 📆  **7/2/18**

# Must
- [x] Introduce Octicons for React #222


lerna.json
```diff
@@ -4,7 +4,8 @@
     "lib/octicons_node",
     "lib/octicons_gem",
     "lib/octicons_helper",
-    "lib/jekyll-octicons"
```
jonrohan: I was thinking this was a good idea too 👍

lerna.json
```diff
@@ -4,7 +4,8 @@
...
+    "lib/octicons_jekyll",
+    "lib/octicons_react"
```
jonrohan: About this list, I did this because they needed to be built in a specific order, `octicons_node` and `octicons_gem` before the other ruby packages. 

I could see moving `octicons_react` up to below `octicons_node` then if rubygems is down, it won't effect our node packages being published.

lib/octicons_react/README.md
```diff
@@ -0,0 +1,200 @@
...
+```jsx
+/**
+ * The prop form is shorter, but doesn't allow you to pass icon props.
+ */
+<Octicon icon={Icon} />
+
+/**
+ * The child form allows you to pass props.
+ */
+<Octicon><Icon x={10}/></Octicon>
```
jonrohan: Is `x={10}` an actual api call? If not, I think we should change it to something like `size='large'` so it doesn't confuse anyone.

If there isn't any props you can pass right now to the child form. What do you think about dropping that example from the readme. I see the value in keeping it, but until we support props to manipulate the icon, there's not much point in promoting it.

lerna.json
shawnbot: Oh, that's good to know! I'm in favor of keeping the ones that don't need to be ordered specifically in alphabetical order.

lib/octicons_react/README.md
shawnbot: Maybe an example that adds a `transform="rotate(180)"`? 🤔 

I guess I'm in favor of removing the docs for that form if we don't use it, but I was kind of expecting us to use it eventually and wanted to keep it open.

jonrohan:
> Assuming it passes CI, does that look good to you?

⚡️ Yeah This is great thanks!

PR #225, opened 2018-07-12
Release v8.0.0

shawnbot:
This is a major version bump because we're changing the name of the [`@github/octicons-react` npm package](https://www.npmjs.com/package/@github/octicons-react) to `@githubprimer/octicons-react` (note that the new scope will be "github**primer**"). We will [deprecate](https://docs.npmjs.com/cli/deprecate) the `github`-scoped package as soon as the `githubprimer`-scoped one is published.

If you're using `@github/octicons-react` and don't want to see deprecation warnings every time you `npm install`, do this (once this PR is merged and the [release](https://github.com/primer/octicons/releases) is posted):

```
npm uninstall --save @github/octicons-react
npm install --save @githubprimer/octicons-react
```

Then find and replace each instance of the old package with the new one, e.g. with:

```sh
perl -pi -e 's#@github/octicons-react#@githubprimer/octicons-react#g' src/**/*.js
```

That should be it! 🤞

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -7,5 +7,5 @@
     "lib/octicons_jekyll",
     "lib/octicons_react"
   ],
-  "version": "7.4.0"
+  "version": "8.0.0"
 }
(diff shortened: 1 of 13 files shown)


emplums:
Published! 🎉 https://www.npmjs.com/package/@githubprimer/octicons-react

PR #228, opened 2018-07-19
Add TypeScript types to the React integration

j-f1:
This adds TypeScript types to `@githubprimer/octicons-react`, which allows TypeScript projects to install the module and immediately start using it without installing a separate package. IMO this package should have the types integrated because they need to be generated while building the package to get the most current list of names to export.

The TypeScript types include the exact values for each icon’s `size` property, and calling `getIconByName` with a literal string representing an icon will return the correct type of that icon.

Ref https://github.com/desktop/desktop/issues/5185.

### DIFF
--- a/lib/octicons_react/package.json
+++ b/lib/octicons_react/package.json
@@ -7,13 +7,14 @@
   "license": "MIT",
   "main": "dist/index.umd.js",
   "module": "dist/index.esm.js",
+  "types": "dist/index.d.ts",
   "repository": "primer/octicons",
   "scripts": {
     "pretest": "npm run lint",
     "test": "jest",
     "start": "NODE_ENV=production next",
     "lint": "eslint src pages script",
-    "prepare": "script/build.js && npm run rollup",
+    "prepare": "script/build.js && script/types.js && npm run rollup",
     "prepublishOnly": "../../script/notify pending",
     "preversion": "npm run prepare",
     "publish": "../../script/notify success",
@@ -28,6 +29,7 @@
     "primer"
   ],
   "dependencies": {
+    "@types/react": "^16.4.6",
     "prop-types": "^15.6.1"
   },
   "devDependencies": {
--- a/lib/octicons_react/script/build.js
+++ b/lib/octicons_react/script/build.js
@@ -5,17 +5,21 @@ const {join, resolve} = require('path')
 
 const srcDir = resolve(__dirname, '../src/__generated__')
 const iconsFile = join(srcDir, 'icons.js')
+const typesFile = join(srcDir, 'icons.d.ts')
 
 function CamelCase(str) {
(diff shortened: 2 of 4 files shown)


j-f1:
> You must set `FIGMA_TOKEN` to a Figma personal access token.

Is there a way to disable that on PR builds?

liamdawson:
Here's a gist showing them in use. I pulled the commit in this PR in and tested it locally, and [made this](https://gist.github.com/liamdawson/17e52e51d59069c72f25eb35949b3e20). (I was testing it in a Storybook).

It'd be a great help if we can get this merged.

liamdawson:
> Also, are there conventions around unit testing TypeScript types? 

Probably define some Typescript examples with strict compiler options and check if the examples compile?

lib/octicons_react/package.json
```diff
@@ -28,6 +29,7 @@
     "primer"
   ],
   "dependencies": {
+    "@types/react": "^16.4.6",
```
shiftkey: I think this can be moved to a `devDependency` as it's not needed at runtime (or for non-TS usage)

shiftkey:
@shawnbot 

> Also, are there conventions around unit testing TypeScript types?

The typical example for JS projects shipping TS declarations is essentially what moment.js has setup:

 - [define a number of tests in a TS file](https://github.com/moment/moment/blob/develop/typing-tests/moment-tests.ts) - these should reflect the real API usage
 - [define a TypeScript config alongside](https://github.com/moment/moment/blob/develop/typing-tests/tsconfig.json) in case you have any unexpected dependencies that need to be included (e.g. using Node APIs)
 - [add a script to your `package.json`](https://github.com/moment/moment/blob/develop/package.json#L88) so you can run it as part of CI or generating releases

lib/octicons_react/package.json
j-f1: It is needed at runtime, though — the `index.d.ts` file directly `import`s it.

j-f1:
I’ll try to get to this today or tomorrow.

lib/octicons_react/package.json
shiftkey: This is how `react-testing-library` is [setup](https://github.com/kentcdodds/react-testing-library/blob/master/package.json#L43-L44) and I didn't have an issue with using that recently in Desktop:

```
  "devDependencies": {
    "@types/react-dom": "^16.0.6",
    ...
```

shiftkey: Also, feel free to relax the version needed here - many packages just do `"@types/react": "*"` unless they depend on specific APIs.

j-f1:
This should be ready to go. Are there any other comments?

jonrohan:
@j-f1, I think it looks good, I'll let @shawnbot take care of it. He's out until next week.

j-f1:
Friendly ping @shawnbot :)

PR #232, opened 2018-08-24
Add arrow icon to set

venetucci:
I added an arrow icon to the set: 

image

### Why?

@mclark is working on a page to show "two dot" comparisons. Early/rough example:

image

Our default PR shows "three dot" comparisons, and @califa recently changed the compare view from showing `...` to an arrow ⬅: 

image

@mclark and I thought having a similar arrow in the "two dot" view might make the most sense, but since this is a direct comparison between two branches, it would point both directions. I adapted the `arrow-small-left` icon. It would look something like this: 

sample direct

### Naming

I wasn't sure if there was a better naming option for this. I went with `arrow-small-both`, although it may be more accurate to use `arrow-small-horizontal`? 

### Sizes

Is it important to do a larger size for this (the original arrow icons have a large and small size), or just stick to the size we'll be using?


jonrohan:
Thanks @venetucci! 

> Is it important to do a larger size for this (the original arrow icons have a large and small size), or just stick to the size we'll be using?

I'm of the opinion that the small arrows should be deprecated. They're not used in github.com and seem redundant given that one could scale down the larger one. This PR  might be a good push for me to move forward with deprecating the small ones.

> I wasn't sure if there was a better naming option for this. I went with arrow-small-both, although it may be more accurate to use arrow-small-horizontal?

I think you should rename to `arrow-both` re implement them based on the larger icon size and not worry about a small version.

venetucci:
@jonrohan Thanks!

I updated that file to use the larger arrow icon as a base, and renamed to `arrow-both`. 

image

jonrohan:
Thanks for the new icon @venetucci, it's going to be included in https://github.com/primer/octicons/pull/238

venetucci:
Woo, thanks @jonrohan! 🎉

unluisco:
I love to read this pull request, shows the kind of companionship that exists in the company.
I would like to work for github some day.

PR #234, opened 2018-08-28
Make Figma API token optional in export script, fix CI issues

### LINKED ISSUE #233: CI builds fail for outside contributors
CI builds are failing for outside contributors because the `FIGMA_TOKEN` environment variable isn't available on fork PR builds. @jonrohan and I discussed the possibility that we might not need the Figma API token anymore, so let's look at that. We only need to read the file and it's public, so maybe?

shawnbot:
This is intended to fix #233 by working around the absence of the `FIGMA_TOKEN` environment variable on Travis. Here's how it works:

* If there is a `FIGMA_TOKEN`, we use that and fetch the icons from Figma as usual.
* Otherwise, we:
    1. Get the `version` field from `lib/octicons_node/package.json`
    1. Fetch the _top-level_ `package.json` from GitHub at the versioned tag (`https://raw.githubusercontent.com/primer/octicons/v${version}/package.json`)
    1. If the `figma.url` field is the same as the one in the local top-level `package.json`, we can assume that the icons haven't changed in this build and fetch the build files from `https://unpkg.com/octicons@${version}/build/`
    1. Otherwise, we die. 💀 

One issue that I ran into getting files from `unpkg.com` was a mysterious SSL/TLS error complaining about a self-signed cert. I _think_ it's the same as [this error](https://github.com/request/request/issues/418), but none of the suggested solutions in that thread worked for me, so I worked around the Node's TLS by shelling out to `curl -sL`. ¯\\\_🤓_/¯


lib/octicons_helper/package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "version": "../../script/rubyversion ./lib/octicons_helper/version.rb",
     "postinstall": "bundle install --path vendor/bundle",
+    "prepare": "echo '(no prepare needed here)'",
```
shawnbot: This makes it easier to just run `npx lerna run prepare` from the top level.

lib/octicons_helper/Gemfile
```diff
@@ -8,6 +8,6 @@ gem "rails"
 group :development, :test do
   gem "minitest"
   gem "rake"
-  gem "rubocop"
-  gem "rubocop-github"
+  gem "rubocop", "0.50"
+  gem "rubocop-github", "0.5.0"
```
shawnbot: I updated these to match the versions in `octicons_gem/Gemfile`.

lib/octicons_react/package.json
```diff
@@ -10,12 +10,14 @@
   "types": "dist/index.d.ts",
   "repository": "primer/octicons",
   "scripts": {
-    "pretest": "npm run lint && npm run ts-test",
+    "build": "script/build.js && script/types.js",
+    "pretest": "npm run lint",
     "ts-test": "tsc -P ts-tests",
     "test": "jest",
+    "posttest": "npm run ts-test",
```
shawnbot: TypeScript tests are run after the main test suite now.

script/try-release
```diff
@@ -3,7 +3,7 @@ set -e
 # pwd
 package=$(jq -r .name package.json)
 version=$(jq -r .version package.json)
-published=$(npm info $package@$version version)
+published=$(npm info "$package@$version" version || echo "0.0.0")
```
shawnbot: This used to throw an error if the package you specified didn't exist. If that's the case, just return `0.0.0`.

Issue #216, opened 2018-04-11
How to use octicons on pure HTML static page?


edpichler:
Can I use these icons on client side, without node.js, jekyl or ruby?
I have a pure HTML page and I would like to use it. Can someone send a link with an example?

shurcooL:
> I have a pure HTML page

Do you mean the HTML is hand-written? It's not generated in any way?

You can do it, you'll just have to manually insert the `<svg>...</svg>` elements into your HTML.

elv1s42:
Hi All,

I am using Octicons on static HTML page too, and it looks like it was easier to use the old version of Octicons (not sure about the exact version number). The structure of Octicons in my project looked like this:

![image](https://user-images.githubusercontent.com/6689884/45021225-abfe6300-b039-11e8-98ff-3519b98b3cd5.png)
As you can see, there were no saparate `.svg` files and I was able to include icons on HTML page like this:
``` html
   <span class="octicon octicon-arrow-right"></span>
```

Looks like now it is not possible, and I need to insert content of each required `.svg` file manually.

Thank you

shawnbot:
Hi everyone,

We currently don't have the need within GitHub to render Octicons outside of Ruby, Node, Jekyll, or React. I'm going to close this issue, but I hope that folks will continue to share links and solutions here because it's clearly important to the broader community. We may revisit this need in the future, but in the meantime I suggest [searching GitHub for open source implementations](https://github.com/search?utf8=%E2%9C%93&q=octicons&ref=simplesearch). (I can't suggest — or link to, generally — any specific ones because we don't have the resources to monitor them for quality and support, but I'm sure that there are good ones out there!)

Thanks for your input and your involvement, and best of luck. :v:

sumeetpareek:
It is actually really easy if you use it this way https://www.youtube.com/watch?v=dO8r6bnc9Q4

anishansari:
> It is actually really easy if you use it this way https://www.youtube.com/watch?v=dO8r6bnc9Q4

How can I use the light version of these icons? is there any class for that ? or how can I make it Light(white) if my background is black? Thanks 
![Capture](https://user-images.githubusercontent.com/4586123/54621829-6e6d9e80-4a8e-11e9-9aca-20af8d4c0b70.PNG)

titomiguelcosta:
> It is actually really easy if you use it this way https://www.youtube.com/watch?v=dO8r6bnc9Q4

[...]

Issue #217, opened 2018-04-17
How to use octicons with Bootstrap 4


blachawk:
Bootstrap claims that your icons work well - https://getbootstrap.com/docs/4.0/extend/icons/

Can you provide any examples of how we would plug your icons into a bootstrap 4 scenario? (i.e. what would the html code look like for a small or very large version of any of your icons on an HTML page with bootstrap 4)?

Thanks

shawnbot:
Hi @blachawk, thanks for this. We don't use Bootstrap at GitHub, so I'm afraid that your question is outside the scope of this repo. See https://github.com/primer/octicons/issues/216#issuecomment-440808401 for more details, and feel free to follow up here with any solutions you've found so that people searching the web can find them if they land here. Cheers!

PR #242, opened 2018-09-06
Octoicons can be easily used by including as script in an html file

shaminmeerankutty:
When looking through octoicons I have found that it is very hard to include octoicons to a simple html file. There is no way without a node or ruby project. So to make it simple I have added an `octoicons_web`. It does have the `octoicons.min.js` in the `dist` folder that can be included to a simple html file and use all those icons in the octoicons by just including the script and an html tag like `<div data-icon="arrow-both"></div>` Also we can specify the width and the height. Now even with a html beginner can use octoicons.


shaminmeerankutty:
This fixes #216,  #217

lib/octoicons_web/src/index.js
```diff
@@ -0,0 +1,36 @@
...
+    "viewBox": "0 0 " + icon_data.width + " " + icon_data.width,
+    "class": "octicon octicon-" + name,
+    "aria-hidden": "true"
+  }
+  icon.insertAdjacentHTML("beforebegin", generateIcon(icon_data))
+  icon.parentNode.removeChild(icon);
+})
+
+function generateIcon(icon_data){
+  var icon =  "<svg " + htmlAttributes(icon_data, icon_data.options) + ">" + icon_data.path + "</svg>"
```
KartoffelCheetah: You are passing icon and icon.options to htmlAttributes.
It is not consistent with htmlAttributes's [definition](https://github.com/primer/octicons/pull/242/commits/fee22773e8591520c37f46922d0f7a1a081d844a#diff-64f0a99e7e5fb7797c348ba46ce3dca8R27).

shaminmeerankutty: @KartoffelCheetah Sorry my bad. I will fix it.

aparcar:
Works here, thanks!

cyberalien:
I offer alternative solution (opened a different issue, so I won't hijack this one): https://github.com/primer/octicons/issues/263

Advantages:
* It is automatically updated from NPM package daily, so it requires no maintenance.
* It downloads only icons that are used on page.
* It is universal solution, not designed specifically for Octicons, so it can be used to combine multiple icon sets on same page: Octicons, FontAwesome, Material Design Icons and others.
* Works with dynamic JavaScript. It observes DOM for changes and loads/renders icons whenever they are added dynamically, such as AJAX forms.

PR #253, opened 2018-11-05
Fix for list-ordered icon

emilybrick:
- Fix list ordered icon in the master Octicons file as per https://github.com/primer/octicons/pull/252
- Updated the changelog


emilybrick:
woooo, ty @shawnbot and @jonrohan !!! ✨

PR #254, opened 2018-11-05
Making some fixes to the export script so that exporting is recursive on the figma doc

jonrohan:
There was some errors happening on exporting in some of the recent builds. I tracked the problem down to the icons being in groups. So I modified the filter function to recursively search for the components.

This fixed the problem 👍

### DIFF
--- a/script/export
+++ b/script/export
@@ -53,10 +53,13 @@ let oCount = 0
 removeSync(outputDir)
 
 // Get the components
-getFigmaComponents()
+getFigmaComponents(figmaFileKey)
   .then(components => {
     // Map all components to a new object
-    return components.reduce((data, c) => {
+    return Object.keys(components).map((k) => {
+
+      const c = components[k]
+      const j = JSON.parse(c.raw)
 
       // Keywords extracted from description when that's ready
       let keywords = (c.description || "").match(/^keywords: (.+)$/mi)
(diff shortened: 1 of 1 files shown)


script/export
```diff
@@ -144,26 +144,31 @@ function getFigmaComponents() {
...
-    .then(components => {
-      oCount = components.length
+      let components = {}
+      const check = (c) => {
+        if (c.type === 'COMPONENT') {
+          components[c.id] = {
+            name: c.name,
+            id: c.id,
+            description: response.body.components[c.id].description,
+            raw: JSON.stringify(c)
```
shawnbot: Why are we stringifying here? Couldn't we do this?


```js
components[c.id] = Object.assign(
  {},
  c,
  response.body.components[c.id]
})
```

script/export
```diff
@@ -144,26 +144,31 @@ function getFigmaComponents() {
...
+          }
+        } else if (c.children) {
+          c.children.forEach(check)
+        }
+      }
+      response.body.document.children.forEach(check)
+      oCount = Object.values(components).length
```
shawnbot:
```suggestion
      const oCount = Object.values(components).length
```

script/export
```diff
@@ -132,10 +131,11 @@ getFigmaComponents()
     process.exitCode = 1
   })
 
-function getFigmaComponents() {
-  spinner.info("Getting components from the figma file")
-  spinner.start(`Contacting ${FIGMA_DOMAIN}`)
-  return got.get(`${FIGMA_DOMAIN}/v1/files/${figmaFileKey}`, {
+function getFigmaComponents(figmaFileKey) {
+  return new Promise((resolve, reject) => {
```
shawnbot: It looks like we don't have to create a new Promise here. I think you can return the `got.get()` promise and just `return components` rather than resolving, then nix the `catch(reject)` bit.

script/export
```diff
@@ -144,26 +144,31 @@ function getFigmaComponents() {
...
+        throw Error("No octicons found.")
+      }
       spinner.succeed(`${oCount} icons found in the figma file`)
-      return components
+      return resolve(components)
     })
     .catch(err => {
-      spinner.fail("Error: Getting getting octicons from figma file")
-      throw err
+      reject(err)
```
shawnbot: Oh, and do we need to stop the spinner here, or is that automatic? If we're unwrapping the `new Promise()`, you can do this:

```js
.catch(err => {
  spinner.stop() // or whatever
  throw err
})
```

PR #270, opened 2018-12-21
Reorganize package directories

### LINKED ISSUE #269: [RFC] Reorganize lib directory
I'd like to propose a reorganization of the `lib` directory to make it easier to find things. Currently, every package is named `lib/octicons_<x>` (which smells bad to me already) and, with one exception, the `<x>` part doesn't match up with any of the published package names. Here's how I think they should be named:

| Before | After |
| :--- | :--- |
| `lib/octicons_gem` | `ruby/octicons` |
| `lib/octicons_helper` | `ruby/octicons_helper` |
| `lib/octicons_jekyll` | `ruby/jekyll-octicons ` |
| `lib/octicons_node` | `node/octicons` |
| `lib/octicons_react` | `node/octicons-react` |

shawnbot:
Fixes #269, and:

* Lints all our JS (scripts and library code!) using the common eslint config.
* Re-scopes `@githubprimer/octicons-react` to `@primer/octicons-react`.
* Re-scopes the dummy npm packages for our Ruby libs to `@primer`.
* Modernizes the JS in `octicons` and adds `"engines": {"node": ">=8"}` to `octicons/package.json` to warn older Node versions that the code may not run (we should compile the browser version with Babel's `env` preset).


node/octicons/index.js
```diff
@@ -0,0 +1,60 @@
...
+    width: icon.width,
+    height: icon.height,
+    viewBox: `0 0 ${icon.width} ${icon.height}`,
+    class: `octicon octicon-${key}`,
+    'aria-hidden': 'true'
+  }
+
+  // Function to return an SVG object
+  icon.toSVG = options => {
+    return `<svg ${icon.getAttributes(options)}>${icon.path}</svg>`
```
colebemis: Does this template literal produce the expected result? Looks like `getAttributes` returns a list and I think lists are joined with `","` not `" 
 "`. 
image

shawnbot: You're right! Guess we don't have tests for this, huh? 😬 
```suggestion
    return `<svg ${icon.getAttributes(options).join(' ')}>${icon.path}</svg>`
```

node/octicons/index.js
```diff
@@ -0,0 +1,60 @@
...
+        attrs['aria-label'] = options['aria-label']
+        attrs.role = 'img'
+
+        // Un-hide the icon
+        delete attrs['aria-hidden']
+      }
+    }
+
+    return Object.keys(attrs).reduce((list, attr) => {
+      list.push(`${attr}="${attrs[attr]}"`)
```
colebemis: Maybe a `map` makes more sense here? Since we're transforming a list into another list. Something like:

```js
Object.entries(attrs).map(([key, value]) => `${key}="${value}"`).join(' ')
```

colebemis: I guess that `join(' ')` could either be done here or in `toSVG` 🤷‍♂️

node/octicons/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+for (const key of Object.keys(icons)) {
+  const icon = icons[key]
+
+  // Set the symbol for easy access
+  icon.symbol = key
+
+  // Set all the default options
+  icon.options = {
+    version: '1.1',
```
colebemis: Is the `version` attribute necessary? Looks like all browsers just ignore it: https://stackoverflow.com/questions/18467982/are-svg-parameters-such-as-xmlns-and-version-needed

node/octicons/index.js
shawnbot: Yep, this reduce is silly. Two things:

1. IIRC `Object.entries()` returns an iterator, so you either need `Array.from()` it or spread it into an array; so maybe just:

    ```js
    Object.keys(attrs).map(key => `${key}="${attrs[key]}"`).join(' ')
    ```

2. We should be escaping attribute values here, otherwise we're opening folks up to all sorts of issues.

colebemis: > 1. IIRC `Object.entries()` returns an iterator, so you either need `Array.from()` it or spread it into an array;

Looks like `Object.entries()` does return an array so `Array.from()` shouldn't be necessary: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

> 2. We should be escaping attribute values here, otherwise we're opening folks up to all sorts of issues.

Good point 👍

node/octicons/index.js
shawnbot: TIL! I'm fine with dropping this; how about you, @jonrohan?

PR #271, opened 2019-01-03
[octicons-react] Add className to the type definition

pocke:
Problem
===

octicons-react's `Octicon` component receives `className` as a prop, but the type definition does not include `className` prop.
https://github.com/primer/octicons/blob/7b5d10a4f3af617084b6bd6bc33700b8e04c2470/lib/octicons_react/src/index.js#L17-L18

Solution
===

Add className to the type definition.


jonrohan:
Thanks for the PR. I'm gonna let @shawnbot take a look at this one.

pocke:
Thank you for reviewing!
Should I do anything for this pull request?

PR #276, opened 2019-01-30
Converting Octicons workflow to a GitHub actions workflow

jonrohan:
This PR will serve as a catch all for a series of pull requests that will convert the octicons/travis etc CI workflow into a GitHub actions workflow.

- [x] https://github.com/primer/octicons/pull/277
- [x] https://github.com/primer/octicons/pull/278
- [x] https://github.com/primer/octicons/pull/279
- [x] https://github.com/primer/octicons/pull/280
- [x] Update contributing docs

cc @shawnbot @emplums


.github/actions/build_node/entrypoint.sh
```diff
@@ -0,0 +1,54 @@
...
+  # Respect NPM_CONFIG_USERCONFIG if it is provided, default to $HOME/.npmrc
+  NPM_CONFIG_USERCONFIG="${NPM_CONFIG_USERCONFIG-"$HOME/.npmrc"}"
+  NPM_REGISTRY_URL="${NPM_REGISTRY_URL-registry.npmjs.org}"
+
+  # Allow registry.npmjs.org to be overridden with an environment variable
+  printf "//%s/:_authToken=%s\\nregistry=%s" "$NPM_REGISTRY_URL" "$NPM_AUTH_TOKEN" "$NPM_REGISTRY_URL" > "$NPM_CONFIG_USERCONFIG"
+  chmod 0600 "$NPM_CONFIG_USERCONFIG"
+fi
+
+PACKAGE_VERSION=$(cat package.json \
```
JasonEtco: There's a library `jq` for getting values from a JSON file:

```suggestion
PACKAGE_VERSION=$(jq '.version' --raw-output ./package.json)
```

You _might_ need to install it in the Dockerfile, but I think its there.

.github/actions/build_node/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM node:10
```
JasonEtco: A `node:10-slim` image will be a lot smaller, so faster to run in Actions. The main difference I know of is that `-slim` doesn't have `git`, so dependencies not from NPM won't install

PR #277, opened 2019-01-30
Creating a version action that versions the release based on PR logic

jonrohan:
This action updates the root `package.json` with a version number based on our primer workflow logic. After updating the version in `package.json` the actions down in the workflow will be able to read this number and publish/update packages.

#### Logic

1. If the pull request branch begins with `release-` then treat that as a release candidate. Grab the version number from the branch name eg `release-8.0.0 => 8.0.0`. The version number will end up with a `rc` postfix and the commit short sha. eg. `8.0.0-rc.1cd4n78`
2. If the pull request isn't `master` and isn't a `release-` branch, then treat this as an alpha release. Using the current version number from package, it will bump it to a patch release and add a postfix of alpha and the commit sha. eg `8.0.0 => 8.0.1-alpha.1cd4n78`

@shawnbot @emplums

### DIFF
--- a/.github/actions/version/entrypoint.js
+++ b/.github/actions/version/entrypoint.js
@@ -1,0 +1,32 @@
+const semver = require('semver')
+const fs = require('fs')
+const {
+  join,
+  resolve
+} = require('path')
+
+ const {
+  GITHUB_REF,
+  GITHUB_SHA
+} = process.env
+const pkg = require(resolve(process.cwd(), 'package.json'))
+const branchName = GITHUB_REF.split('/').pop()
+
+ const writePackageJson = () => {
+  fs.writeFileSync(join(process.cwd(), 'package.json'), JSON.stringify(pkg), 'utf8')
+}
+
+ // If it's a release branch
+if (branchName.match(/^release-/i)) {
+  console.log('Versioning release candidate')
+  const newVersion = branchName.match(/^release-([\d\.]+)/i)
+  pkg.version = `${newVersion[1]}-rc.${GITHUB_SHA.slice(0,7)}`
+  writePackageJson()
+}
+
+ // Otherwise
+else if (branchName != 'master') {
+  console.log('Versioning prerelease')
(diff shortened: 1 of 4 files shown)


.github/actions/version/entrypoint.js
```diff
@@ -0,0 +1,32 @@
...
+} = process.env
+
+const pkg = require(resolve(process.cwd(), 'package.json'))
+const branchName = GITHUB_REF.replace(/^refs\/(heads|remotes\/origins)\//gi, "")
+
+const writePackageJson = () => {
+  fs.writeFileSync(join(process.cwd(), 'package.json'), JSON.stringify(pkg), 'utf8')
+}
+
+ // If it's a release branch
```
shawnbot:
```suggestion
// If it's a release branch
```

.github/actions/version/entrypoint.js
```diff
@@ -0,0 +1,32 @@
...
+  join,
+  resolve
+} = require('path')
+ const {
+  GITHUB_REF,
+  GITHUB_SHA
+} = process.env
+
+const pkg = require(resolve(process.cwd(), 'package.json'))
+const branchName = GITHUB_REF.replace(/^refs\/(heads|remotes\/origins)\//gi, "")
```
shawnbot: We should double-check with the Actions team if `GITHUB_REF` ever begins with anything other than `refs/heads/`. If not, I think we can safely remove the `|remotes\/origins` bit.

.github/actions/version/entrypoint.js
```diff
@@ -0,0 +1,32 @@
...
+
+const writePackageJson = () => {
+  fs.writeFileSync(join(process.cwd(), 'package.json'), JSON.stringify(pkg), 'utf8')
```
shawnbot: I _think_ that prefixing the `package.json` path with the cwd is redundant here. Also, we should pretty-print the JSON back:

```suggestion
  fs.writeFileSync('package.json', JSON.stringify(pkg, null, 2), 'utf8')
```

PR #279, opened 2019-01-31
Creating a custom action for building all javascript based octicons packages

jonrohan:
This is a common action that builds any npm based project. `octicons_node` and `octicons_react`

The steps it does is 

1. change to the directory of the project
2. Install
3. Build
4. Lint
5. Test

@shawnbot @emplums


lib/octicons_node/script/build.js
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
+
+console.log("Create a build/data.json from the exported assets")
```
jonrohan: TODO: Right now the figma-action is specific to octicons and is still building the build/data.json for us. I'd like to make it more generic and do the conversion work here. Until then

lib/octicons_node/package.json
```diff
@@ -17,9 +17,8 @@
     "url": "https://github.com/primer/octicons/issues"
   },
   "scripts": {
-    "publish": "../../script/notify success",
```
emplums: Should we keep these in here until the publish stuff is merged in?

jonrohan: Actions notifies for us now. 🥇

PR #281, opened 2019-02-06
v8.4.0

jonrohan:
Release v8.4.0

### :house: Internal

- [x] Using Actions to build and deploy Octicons https://github.com/primer/octicons/pull/276

--- 
Release Steps
- [x] `npm version 8.4.0` on this branch


.github/CONTRIBUTING.md
```diff
@@ -59,9 +59,11 @@ Once the new changes have been added the new components to the master Octicons f
 
 After you've published the latest changes in the Figma file, create a release branch in the [Octicons repo](https://github.com/primer/octicons).
 
-Update the [changelog](https://github.com/primer/octicons/blob/master/CHANGELOG.md) with the new changes pertaining to this version.
+Update the [CHANGELOG](https://github.com/primer/octicons/blob/master/CHANGELOG.md) with the new changes pertaining to this version.
 
-Once the changlog has been updated, run `npm run bump`. This will guide you through a prompt asking what all the new versions of the packages should be. In the context of Octicons, significant changes to the library or workflow, or removing an icon would be considered a major update, adding a new icon would be considered a minor update, and fixing an icon would be considered a patch. Reach out in the #design-systems Slack channel if you're unsure! 
+Once the CHANGELOG has been updated, run `npm version <newversion>`. This will update package.json with the new version, then update all the `lib/*` packages with the same version. If that runs smoothly, it should commit the changed files. Push that commit to your release branch.
```
shawnbot: One thing to note here: we might want to add `git-tag-version = false` to our (non-existent) `.npmrc` so that running this doesn't "pollute" your local tags with unreleased commits.

.github/actions/build_node/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:10-slim
+
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y \
+        jq && \
+        apt-get clean -y && \
+    rm -rf /var/lib/apt/lists/*
```
shawnbot: What's this last `rm -rf` bit do? Is that to make the container images smaller?

jonrohan: Not sure, I copied it from https://github.com/actions/bin/blob/9d4ef995a71b0771f438dd7438851858f4a55d0c/filter/Dockerfile#L14

.github/actions/version/entrypoint.js
```diff
@@ -0,0 +1,37 @@
...
+ const {
+  GITHUB_REF,
+  GITHUB_SHA
+} = process.env
+
+// Octicons package
+const pkg = require(resolve(process.cwd(), 'package.json'))
+
+// GitHub info
+const branchName = GITHUB_REF.replace(/^refs\/heads\//gi, "")
```
shawnbot: Nit: since `refs/heads/` should only exist once, you should be able to get away with a string replacement, which is more legible:

```suggestion
const branchName = GITHUB_REF.replace('refs/heads/', '')
```

.github/actions/version/entrypoint.js
```diff
@@ -0,0 +1,37 @@
...
+const shortSha = GITHUB_SHA.slice(0,9)
```
shawnbot: For what it's worth, I chose 7 in primer/publish because we're _probably_ not going to run into any SHA collisions in our projects. Do think think that's a risk here, or is 7 safe?

jonrohan: 7 is fine with me

.github/CONTRIBUTING.md
jonrohan: But I do want npm to commit the change. Might leave it as is for now

lib/octicons_node/package.json
```diff
@@ -17,9 +17,8 @@
     "url": "https://github.com/primer/octicons/issues"
   },
   "scripts": {
-    "publish": "../../script/notify success",
-    "prepublishOnly": "../../script/notify pending",
-    "prepare": "cp -R ../build . && cp index.scss build/build.css",
+    "build": "script/build.js",
+    "lint": "eslint index.js tests/*.js",
```
shawnbot: Not a blocker, but could we also lint the `script` directory since some of those are JS?

lib/octicons_node/package.json
```diff
@@ -30,5 +29,10 @@
   ],
   "dependencies": {
     "object-assign": "^4.1.1"
```
shawnbot: Not for now, but we should look into removing this dependency in favor of `Object.assign()`, since we're very likely not interested in supporting browsers that don't have it anymore.

lib/octicons_react/package.json
```diff
@@ -45,7 +41,6 @@
     "fs-extra": "^6.0.1",
     "jest": "^23.2.0",
     "next": "^5.1.0",
-    "octicons": "8.3.0",
     "primer-react": "0.0.6-alpha.1",
```
shawnbot: 🙈 /cc @emplums

jonrohan: never have dependencies, that's what I say

.github/CONTRIBUTING.md
shawnbot: Oh right, yeah — it's weird that with that option, npm skips the commit too. Disregard!

lib/octicons_node/package.json
jonrohan: for now, script/build is a todo. there's nothing there to lint

jonrohan:
> The only thing I'm not 100% sure of is whether this changes whether/where the JSON data files live in the published npm packages. If there's any possibility of "breaking" that API, we might want to make this a major version bump. Otherwise, 🚀

For now it's the same, generated by the figma action, however I would like to have the figma action be more generic and then build the data.json based on output from it

PR #295, opened 2019-02-22
Fix octicons_gem including aria-hidden when aria-label is present

muan:
Octicons gem should check for keys of both symbol and string types.

---

I was unable to run the test locally because it is looking for a mysterious `./build/data.json` file, and I can't find instructions for building one locally. 🤔

### DIFF
--- a/lib/octicons_gem/lib/octicons/octicon.rb
+++ b/lib/octicons_gem/lib/octicons/octicon.rb
@@ -43,8 +43,8 @@ module Octicons
     def a11y
       accessible = {}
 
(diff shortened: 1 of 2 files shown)


jonrohan:
> I was unable to run the test locally because it is looking for a mysterious `./build/data.json` file, and I can't find instructions for building one locally.

sorry about that. As it stands now, it requires setting up figma api keys to be able to build the data file locally. I'm hoping to add some extra to the code that will pull the last release file data.json if there's no figma api token in someone's env.

Otherwise this looks like a good change! I can get it out next week.

PR #310, opened 2019-05-23
added shield icon +states for security advisories

ashygee:
PR to add @donokuda's shield icons for Security Advisories as seen in https://github.com/github/github/pull/114086


jonrohan:
Shipped!

PR #311, opened 2019-05-23
Scoping npm packages to organization @primer/octicons  and @primer/octicons-react

jonrohan:
This updates both the npm octicons package and the npm octicons-react package to live under the `@primer` org name.

@shawnbot


emplums:
Are there any other documentation places like primer.style that should be updated too?

jonrohan:
> Are there any other documentation places like primer.style that should be updated too?

From what I can tell, everything else is import related https://github.com/search?q=org%3Aprimer+%22githubprimer%2F%22&type=Code

PR #322, opened 2019-06-20
Test for duplicate icon names and fail when two icons have the same name

jonrohan:
Caveat: This isn't currently failing because the data being output needs massaging. We're [currently using a sha'd version of the exporter](https://github.com/primer/octicons/blob/030a66ff887d8462f8f8ec30ca1e0344e7b09dc9/.github/main.workflow#L36) because I haven't had time to adjust the build processes for data structure changes. 

When this does work it will go through each icon and test if there's another icon with the same name but  different figma id. When that happens it will fail.

### DIFF
--- a/tests/index.js
+++ b/tests/index.js
@@ -5,6 +5,7 @@ const globby = require('globby')
 const year = new Date().getFullYear()
 const yearRegex = new RegExp(`Copyright \\(c\\) ${year} GitHub Inc\\.`)
 const octiconsLib = fs.readdirSync('./lib/build/svg')
+const octiconsData = require('../lib/build/data.json')
 
 test(`LICENSE files have the current year ${year}`, t => {
   return globby(['**/LICENSE', '!**/node_modules/**/LICENSE', '!**/vendor/**/LICENSE']).then(paths => {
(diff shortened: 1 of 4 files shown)


tests/index.js
```diff
@@ -19,3 +20,19 @@ test(`LICENSE files have the current year ${year}`, t => {
 test('SVG icons exist', t => {
   t.not(octiconsLib.length, 0, `We didn't find any svg files`)
 })
+
+test('Data file exist', t => {
+  t.not(octiconsData.length, 0, `We didn't find any data files`)
+})
+
+const names = {}
+for (const o of Object.values(octiconsData)) {
```
shawnbot: Nit: single-character variable should be `octicon`, `icon`, etc.

tests/index.js
```diff
@@ -19,3 +20,19 @@ test(`LICENSE files have the current year ${year}`, t => {
...
+  test(`No duplicate ${o.name} icon`, t => {
+    if (names[o.name]) {
+      t.fail(`Found duplicate '${o.name}' icons in the figma file. Please rename one of them.`)
```
shawnbot: We could output more information in this failure message, e.g.

```js
t.fail(`Found duplicate '${o.name}` icon: first found was: ${JSON.stringify(names[o.name], null, 2)}`)
```

PR #326, opened 2019-06-27
Adding headphones and empty heart

askdesigners:
This PR adds a headphones icon, and an empty heart.

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -38,7 +38,7 @@ action "Export SVG from Figma" {
     "FIGMA_TOKEN"
   ]
   env = {
-    "FIGMA_FILE_URL" = "https://www.figma.com/file/FP7lqd1V00LUaT5zvdklkkZr/Octicons"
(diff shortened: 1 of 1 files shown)


ashygee:
Hey @askdesigners! Thanks for your contribution. We are in the process of updating our contribution guidelines to limit to just GitHub staff but it just so happens that our team was planning on creating an outlined heart octicon so good timing! 😊We would be happy to include your the `outline-heart` icon with some minor changes but unfortunately cannot include the `headphones` icon as it isn't a part of our UI. 

If you can outline the stroke so it's contained in one single path, I can go ahead and include the `outline-heart` in our next release.

askdesigners:
Hi @ashygee Ok great! I will do that today. Sorry to hear that about the headphones though :/ We are building an extension for VSCode (codetime) and currently have to use an ugly emoji headphones. We were hoping to get a snazzy octicon instead. :) 

So the icons are limited strictly to what is used on Github? Thanks for the message!

askdesigners:
OK @ashygee I have updated the figma with the outlined shape. 👍

PR #335, opened 2019-08-26
[React RFC] Create a separate file per icon

mxstbr:
**Background**

In an internal project (will send specifics in Slack) we are running into hefty bundle size problems caused by @primer/octicons-react bundling all icons into a single file. webpack tree shaking does not recognize that most of the icons are unused, and bundles them all (very similar to [our @primer/components woes](https://github.com/primer/components/issues/456)).

After trying to work around this inside of our project I realised that it's an impossible task, as some @primer/components components import the octicons-react package, and I cannot remove those dependencies. Thusly this RFC to fix this issue upstream.

----

**Summary**

This RFC adds a build step to `octicons-react` based on [svgr](https://github.com/smooth-code/svgr) (which uses svgo under the hood), which creates a separate JS file per icon. For example:

```SVG
<!-- code.svg -->
<svg xmlns="http://www.w3.org/2000/svg" width="14" height="16" viewBox="0 0 14 16">
  <path fill-rule="evenodd" d="M9.5 3L8 4.5 11.5 8 8 11.5 9.5 13 14 8 9.5 3zm-5 0L0 8l4.5 5L6 11.5 2.5 8 6 4.5 4.5 3z"/>
</svg>
```

is turned into

```JS
// Code.js
import React from 'react'

const SvgCode = props => (
  <>
    <path
      fillRule="evenodd"
      d="M9.5 3L8 4.5 11.5 8 8 11.5 9.5 13 14 8 9.5 3zm-5 0L0 8l4.5 5L6 11.5 2.5 8 6 4.5 4.5 3z"
    />
  </>
)

export default SvgCode
```

For users of @primer/octicons-react, the usage would change as follows:

```JS
/*
 * Before
 */
import Octicon, { Code } from '@primer/octicons-react';

/*
 * After
 * Note that the main @primer/octicons-react import contains only the <Octicon /> component, nothing else
 */ 
import Octicon from '@primer/octicons-react';
import Code from '@primer/octicons-react/icons/Code';
```

This is a slightly worse developer experience, but the user experience benefits are massive: this would slim our JS bundle by ~20% (~70kB min+gzip).

----

Note that this RFC is incomplete, but hopefully enough to start a discussion. If there is consensus that this change should be merged, I will finish the implementation.

What are your thoughts?


mxstbr:
Yeah that makes a lot of sense to me, I prefer that!

If that signifies agreement about the general idea I would be happy to finish a first version of this tomorrow, including wrapping each icon automatically.

mxstbr:
Awesome! I just cleaned this up, it's ready for a first review. Here's what the generated component files look like now:

```JS
import React from 'react'
import Octicon from '@primer/octicons-react'

const SvgAlert = () => (
  <>
    <path
      fillRule="evenodd"
      d="M8.893 1.5c-.183-.31-.52-.5-.887-.5s-.703.19-.886.5L.138 13.499a.98.98 0 000 1.001c.193.31.53.501.886.501h13.964c.367 0 .704-.19.877-.5a1.03 1.03 0 00.01-1.002L8.893 1.5zm.133 11.497H6.987v-2.003h2.039v2.003zm0-3.004H6.987V5.987h2.039v4.006z"
    />
  </>
)

SvgAlert.size = [16, 16]

const Alert = props => (
  <Octicon {...props}>
    <SvgAlert />
  </Octicon>
)

export default Alert
```

By importing the `Octicon` component from the root directly we should avoid any duplication of that code. Also note how it's split into two components to make sure `Icon.size` is handled.

**Open questions**

- I am not sure how to handle the UMD bundle, as it should still contain all icons. Any ideas how we are going to handle that? Right now it only contains the `Octicon`.

mxstbr:
I did it! :tada: :tada: :tada: There is now no duplication, the entire build process only generates the icons once.

Here is an overview of the steps in the React build process now:

1. **Compile `.svg` icons to React components**: `svgr` turns all our SVG icons into uncompiled React components in `src/icons/`.
2. **Compile icon components with Babel**: Babel compiles our icon components from `src/icons/` to the root of the package to allow for nice imports (`import Bla from '@primer/octicons-react/Bla'`)
3. **Compile icons.js file with all icons**: The custom build script uses the list of icons to create `src/__generated__/icons.js`, which now _imports_ (rather than inlines) all icons from `src/icons/` (the uncompiled React components)
4. **Rollup compiles the ESM and UMD bundles**: These now have separate entry points and build processes. The ESM bundle (in `src/index.js`) only contains the `Octicon` component, nothing else. The UMD bundle (in `src/index.umd.js`) exports everything from the ESM bundle, but then also contains and exports everything from the `src/__generated__/icons.js` file—which contains all the icons!

Essentially, I managed to get rid of the duplication of the custom build script also turning the `.svg` icons into React components. I hope that makes sense, I feel okay about this whole build process now! Any thoughts?

mxstbr:
The only thing I haven't been able to figure out is TypeScript, for some reason I can't seem to type any of the icon components 🤔

mxstbr:
I opened #345 and now actions run here? 🤔🤔🤔🤔🤔

mxstbr:
Just one tiny piece is missing, I have to manage to type the `src` folder then we should be good to go! :scream:

Sfancy44:
I am not able to do the work at this time please assign someone else to the
task

mxstbr:
Oh my god it finally passed. @shawnbot this is ready for another review, including TS support! :tada:

emplums:
Chatted with Shawn & I'll do some testing with the release candidate this week before we merge this! :)

emplums:
@colebemis also mentioned that this should include bumping the version a major version as well since it's a breaking change

mxstbr:
Yep this should definitely be a new major version. Let me know if you run into any troubles! :pray:

mxstbr:
Friendly ping, any thoughts on this PR? I'd love to move forward with this!

mxstbr:
@shawnbot let me know if there is anything I can do to move this along, we would really like to ship this patch :blush:

lib/octicons_react/pages/index.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import {Block, Text} from 'primer-react'
 import pkg from '../package.json'
-import Octicon, {iconsByName} from '../'
+import Octicon, {iconsByName} from '../index.umd'
```
emplums: I thought we needed to remove `iconsByName`?

emplums:
@mxstbr sorry for the delay on this, I've had a ton on my plate the last few weeks! I'm just getting around to testing out the release candidate now on @primer/components. 

I noticed one thing - I've switched out the `Check` component being used in the `CircleOcticon` example in the docs site with the proper import path, and didn't make any other changes to the `CircleOcticon` component and now I'm seeing an alignment issue:

**Before:**
image
image



**After:**
image
image

emplums:
I'm also noticing that the `Octicon` export doesn't seem to be working properly?
This is in my tests:
```
import Octicon from '@primer/octicons-react'
import PropTypes from 'prop-types'
import styled from 'styled-components'
import {COMMON} from './constants'
import theme from './theme'

const StyledOcticon = styled(Octicon)(COMMON)

StyledOcticon.defaultProps = {
  theme
}

StyledOcticon.propTypes = {
  ...COMMON.propTypes,
  theme: PropTypes.object
}

export default StyledOcticon
```

and I'm getting the error: 

```
Summary of all failing tests
 FAIL  src/__tests__/StyledOcticon.js
  ● Test suite failed to run

    Cannot create styled-component for component: undefined.
```

lib/octicons_react/README.md
```diff
@@ -10,100 +10,34 @@ $ npm install @primer/octicons-react --save
...
-        <li key={key}>
-          <tt>{key}</tt>
-          <Octicon icon={iconsByName[key]}/>
-        </li>
-      ))}
-    </ul>
-  )
-}
-```
+### `<Octicon>`
```
emplums: Does this mean we don't need to wrap components in `<Octicon>` anymore?

emplums: For example - how would this component be rewritten with the new structure? https://github.com/primer/components/blob/master/src/StyledOcticon.js

emplums: I've tried:
```
function StyledOcticon({icon: Icon, size, ...rest}) {
  return (
    <Box {...rest}>
      <Icon size={size} {...rest} />
    </Box>
  )
}
```
but the color and background props don't seem to pass through to the icon

lib/octicons_react/pages/index.js
mxstbr: It is still there in the UMD bundle, since that contains all the icons anyway we might as well keep it around! :blush: I can remove it if you prefer though.

mxstbr:
As you correctly said, the Octicon component is no longer a thing, every icon is automatically wrapped. I will check out that size issue, great catch! 👍 

Regarding the `StyledOcticon`, your implementation honestly looks like it should work, I will check that out too.

colebemis:
Closing this because tree-shaking should achieve the same result.

PR #347, opened 2019-09-09
Update gemspec to allow for Jekyll 4.0

ntotten:
Jekyll 4.0 seems to work fine with this plugin, but the spec doesnt allow it.


jonrohan:
This seems ok, but I want to test it by pulling it into a release pr to make sure things still work.

PR #354, opened 2019-10-17
[React] Allow 'unset' value for verticalAlign property

Fs00:
This PR adjusts type-checking of the React Octicon component to allow users to set the `verticalAlign` property to `unset`.
I'm currently using Octicons for a project of mine and without this change I can't perfectly center an octicon vertically inside a flex container, since `vertical-align` CSS property on the icon clashes with flex `align-items`.


jonrohan:
Thanks for the contribution @Fs00 I'm going to add it to 9.3.0

Fs00:
Thanks!

Issue #349, opened 2019-09-23
Proposal: Infinity and line-arrow icons


superbryntendo:
GitHub uses an infinity icon and a line-arrow on the pricing page and we'd like to use them in signup. Unfortunately, it appears that marketing had to use a non-octicons set to make those happen?

I've drafted up a couple icons to achieve the same stylistic goals while matching the octicons checkmark and x which are the most-used icons on these pages (provided for context below).

![image](https://user-images.githubusercontent.com/4926824/65466424-a1189b00-de13-11e9-8f09-fcd0f730519c.png)

The line-arrow icon is derived from the chevron octicon and is already in use in a signup flow experiment and I'd like to use the infinity icon in another one shortly. I think this would be a particularly nice addition especially since it'd let us support all of the cases on the pricing page too!

yaili:
This has been done.

PR #365, opened 2020-01-07
Adds infinity, line-arrow, and heart-outline symbols, cleans up heart icon path

superbryntendo:
title | image | description
-----------|---|---
`heart` glyph updates | image | Outlining the heart in the glyph vs via css enables the stroke to be centered on the path without increasing the glyph frame dimensions. Regardless of stroking method, the inset point needs to be slightly rounded or the angle increases in an awkward manner. This also updates the vertical centering in the frame to make it easier to align with text.
`heart` glyph diff detail | image | There were a few unnecessary points on the path with some misaligned beziers that made it slightly bumpy, rather than smooth.
`infinity` glyph ([proposal/rationale](https://github.com/primer/octicons/issues/349)) | image | This is for use on pricing pages for features marked as "unlimited" (e.g. contributors, repos, etc.)
`line-arrow` glyphs | image | These are based on the caret/chevron symbols and are used on pricing pages to denote "<- contains all of these features". We also use the up arrow on small screens when the plans reorient to be vertically-listed.

cc @katmeister

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Export SVG from Figma
       uses: primer/figma-action@d7844a1927da9b8dd562cbf008c7da20e228607d
       env:
-        FIGMA_FILE_URL: https://www.figma.com/file/FP7lqd1V00LUaT5zvdklkkZr/Octicons
+        FIGMA_FILE_URL: https://www.figma.com/file/KeVvQazRDZKeTmT9By0bnZ/Octicons-Bryn-s-changes?node-id=0%3A1
         FIGMA_TOKEN: ${{ secrets.FIGMA_TOKEN }}
       with:
         args: format=svg dir=./lib/build


ashygee:
HI @superbryntendo sorry for the delay on this. Just wanted to get a few more details before we put these in the next release.

A few things to note before we put in these updates:
1. the sponsors hearts will need to be updated to use 2 separate icons instead of just CSS. @katmeister do you know who would be best to make aware of this change after these new octicons go out?
2. would you be able to swap the naming convention to prepare for the future update? (`heart/heart-fill` instead of `heart/heart-outline`)

For the addition of the `infinity` and `line-arrows` who should we notify to swap out the SVGs with the octicons after the release?

superbryntendo:
I've changed the `heart` naming scheme.

Can't mention github teams directly here, but pe-nux manages the pages they're implemented on now. site-design has used a 3rd-party icon set to get these glyphs for the pricing page. Hopefully this should allow them to switch over as well.

katmeister:
> 1. do you know who would be best to make aware of this change after these new octicons go out?

@ashygee That would be me! 😄 

> 2\. would you be able to swap the naming convention to prepare for the future update? (`heart/heart-fill` instead of `heart/heart-outline`)

Ah, so there will be a separate icon for the outlined heart? Sure, I can prepare for that. What's the timeline for the next release?

ashygee:
> Ah, so there will be a separate icon for the outlined heart? Sure, I can prepare for that. What's the timeline for the next release?

I will work with @jonrohan together at the end of the week and we should be able to get this set for next week.

jonrohan:
Going in release 9.4.0 https://github.com/primer/octicons/pull/368

PR #373, opened 2020-02-22
Fixes failing typed component

jeremygottfried:
The function fails in practice:
ReactElement is the correct typing for returning a react element.
```sh
JSX element type 'ReactNode' is not a constructor function for JSX elements.
```

`icon` should be left out of props here because this component overrides `icon` prop by default.
It also fails in practice because `OcticonProps` requires `icon`.
```sh
Property 'icon' is missing in type props but required in type OcticonProps
```
Another option is pass `OcticonProps` as a partial to make `icon` optional. `Partial<OcticonProps>`

### DIFF
--- a/lib/octicons_react/ts-tests/index.tsx
+++ b/lib/octicons_react/ts-tests/index.tsx
@@ -15,6 +15,8 @@ function Icon({boom}: {boom: boolean}): React.ReactNode {
   return <Octicon icon={boom ? Zap : Beaker} />
 }
 
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
+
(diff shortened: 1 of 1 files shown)


jonrohan:
I want to hear @colebemis's review on this, since his typescript skills are better.

colebemis:
I’ll look this over this afternoon.

jeremygottfried:
@colebemis Awesome! I might have time to add a test tonight. There may be similar issues in some of the other components in this file, so I'll look into it a bit more.

colebemis:
Checks weren't running because this PR was created from a fork, but I checked out the branch locally and tested myself. Everything looks good  ✅ 

This change doesn't require a version bump since it's only affecting a test file.

Btw, it looks like our GitHub Actions workflow doesn't even run the TypeScript tests 😅 @jonrohan

colebemis:
Thanks again for the contribution, @jeremygottfried! 🎉

Issue #372, opened 2020-02-20
[RFC] Moving the Octicons source of truth into GitHub


colebemis:
## TL;DR

I’m proposing that we move the Octicons source of truth from Figma into GitHub because it would make icon changes easier to track, preview, and publish. 

## Problem

We currently have a [master Figma file](https://www.figma.com/file/FP7lqd1V00LUaT5zvdklkkZr/Octicons?node-id=24251%3A2) that is the source of truth for Octicons. When we push a commit to the master branch of the [primer/octicons](https://github.com/primer/octicons) repo, a GitHub action exports SVG data from the master Figma file and uses that data to build various packages. Here’s a diagram of the current setup:

![Source of truth](https://user-images.githubusercontent.com/4608155/74988416-a265e800-53f2-11ea-8a9e-cba4b1734bf0.png)

> You can learn more about how we currently use Figma in [this blog post](https://github.blog/2018-04-12-driving-changes-from-designs/) by @jonrohan. 

This setup is nice because contributors don’t need experience with git or a local development setup to contribute. However, there are a few drawbacks:

### 1. Icons are not version controlled 
It’s difficult to see how an icon has changed over time or recover an icon if it accidentally gets deleted. Figma has a “Version History” feature but it’s capabilities are pretty limited compared to git.

### 2.  Mistakes can easily go unnoticed
Changes to the master Figma file don't go through the same review process that code does. To illustrate the issue, let’s imagine that a maintainer accidentally deletes an icon while updating the master Figma file. The next time someone pushes a commit to the master branch of [primer/octicons](https://github.com/primer/octicons) that icon would be removed from all the packages without anyone noticing. 

### 3. Contributors don’t get recognized
Here’s how contributions currently work: A contributor begins the process by opening an issue with a link to a Figma file that contains the icon they’d like to contribute. If approved, a maintainer copies the SVG from the given Figma file into the master Figma file and closes the issue. Since no code was merged into the repo, the contributor is not recognized in the “contributors” list on the GitHub repo. 

### 4. It’s hard to preview changes
We currently rely on before/after screenshots and Figma files to preview icon changes. However, it’s hard to understand the impact of a change without seeing the icon used in context.

### 5. Releasing new icons requires many manual steps for maintainers
[...]

PR #374, opened 2020-02-26
Move source of truth from Figma into GitHub

colebemis:
## TL;DR

I moved the Octicons source of truth from Figma into GitHub to make icon changes easier to track, preview, and publish. 

## Problem

We currently have a [master Figma file](https://www.figma.com/file/FP7lqd1V00LUaT5zvdklkkZr/Octicons?node-id=24251%3A2) that is the source of truth for Octicons. When we push a commit to the master branch of the [primer/octicons](https://github.com/primer/octicons) repo, a GitHub action exports SVG data from the master Figma file and uses that data to build various packages. Here’s a diagram of the current setup:

image

This setup is nice because contributors don’t need experience with git or a local development setup to contribute. However, there are a few drawbacks:

1. Icons are not version controlled 
2.  Mistakes can easily go unnoticed
3. Contributors don’t get recognized
4. It’s hard to preview changes
5. Releasing new icons requires many manual steps for maintainers

👉  See https://github.com/primer/octicons/issues/372 for a more detailed explanation of the problems listed above.

## Solution

In this pull request, I moved the source of truth from Figma into GitHub. The `.svg` files for each icon now live in an `/icons` directory in this repo and are used to build the various packages. Here’s a diagram of how this new setup works:

image

With this setup, icon changes are easier to track, preview, and publish.

👉  See https://github.com/primer/octicons/issues/372 to learn more about why this new setup is better than our old one.

## Implementation

Here are the steps I took to move the source of truth from Figma into GitHub:

* Created an top-level `icons` directory that contains `.svg` files for every icon.
* Wrote a [`build.js`](https://github.com/primer/octicons/pull/374/files#diff-4c7b78df589b45477fbf98b2ea9862ae) script to generate a `data.json` file that contains information about the all icons in the `icons` directory.
* Added a [`build`](https://github.com/primer/octicons/pull/374/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R8) npm script (`npm run build`) to make it easier to run the `build.js` script. 
* Updated the [`push.yml`](https://github.com/primer/octicons/pull/374/files#diff-d9e8161f95fdd9b8e476fd456f35a9a2) workflow to run `npm run build` instead of the Figma action.
* Updated `CONTRIBUTION.md` to reflect the new contribution workflow
* Replaced the "icon submission" issue template with an "icon request" template because icon submissions will now be made through pull requests instead of issues.

## Code review

This pull request adds *a lot* of `.svg` files. To make code review easier, I'd recommend filtering out `.svg` files in the "File Changed" tab:

image

## Questions

* Should this pull request be merged into the `v2` or `master` branch? There is nothing v2-specific about these changes except for the `.svg` files, which can easily be swapped out. Is there any advantage to merging into `master` instead of `v2`?

Closes https://github.com/primer/octicons/issues/372

## TODO
- [x] Handle keywords
- [x] Handle invalid input in `build.js` script
- [x] Write tests for `build.js` script
- [x] Update documentation


lib/octicons_helper/test/octicons_helper_test.rb
```diff
@@ -7,7 +7,7 @@
     end
 
     it "renders the svg" do
-      assert_match /<svg.*octicon-x.*><path.*\/><\/svg>/, octicon_v2("x")
+      assert_match /<svg.*octicon-x.*>.*<\/svg>/, octicon_v2("x")
```
colebemis: This test was failing because the new generated SVGs use a non-self-closing `<path>` element (`<path></path>` instead of `<path />`). We probably shouldn't require that SVGs only have a single self-closing `<path>`. I updated the test to allow anything between the opening and close `<svg>` tags.

colebemis:
@yaili I'd love to get your feedback on the changes to the contribution guidelines 🙂 https://github.com/primer/octicons/pull/374/files#diff-8cc7b2b0d78dd2501610391c086a8516

joelhawksley:
@colebemis IMO, this work should be done against `master`. That way we can reduce the surface area of the set of changes in `v2`.

tests/build.js
```diff
@@ -0,0 +1,48 @@
...
+  )
+
+  const {stdout} = await execa(path.resolve(__dirname, '../script/build.js'), ['--input=../icons/**/*.svg'], {
+    cwd: __dirname
+  })
+
+  const icons = JSON.parse(stdout)
+
+  t.true(schema.isValidSync(Object.values(icons)))
+})
```
colebemis: If anyone knows of a better way to test the output of the `build` script, I'd love to hear it 😸

colebemis:
Phew, okay, the commit history is a hot mess but this branch should be good to merge into `master` now ✅ 

cc @joelhawksley

joelhawksley:
@colebemis I'm going to defer to @jonrohan for the review here as he has the historical context on this build system

.github/workflows/push.yml
```diff
@@ -4,26 +4,21 @@ jobs:
...
-        path: ./lib/build
+      - uses: actions/checkout@master
+      - uses: actions/setup-node@master
+        with:
+          node-version: '10.x'
+      - name: Version
+        uses: ./.github/actions/version
+      - run: npm install
+      - run: npm run build
+      - run: cp -r icons lib/build/svg
```
jonrohan: Is it necessary to copy the icons into the build folder anymore? If we are reading from data.json in all the packages we might as well keep the folder out. Probably save in package size.

colebemis: I did this to pass tests but I agree that it seems unnecessary. However, removing the `svg` directory would technically be a breaking change which I was trying to avoid in this PR. Maybe we can remove the `svg` directory as part of the Octicons v2 release since that will also require a major version bump. How does that sound?

PR #375, opened 2020-03-03
Add `internal-repo` icon

ashygee:
This PR addresses the missing octicon for internal repos noted in https://github.com/primer/octicons-v2/issues/300.

cc @jasonlong @cegal

PR #378, opened 2020-03-04
Release 9.5.0

colebemis:
## Features
- New icon `internal-repo` https://github.com/primer/octicons/pull/375


ashygee:
> Doesn't affect this pr, but #375 should have merged into this pr instead of master. I don't think it'll cause a version mismatch.

yeah that was my fault. will be sure to do that next time!

PR #380, opened 2020-03-06
add new north-star icon

ashygee:
adding a new icon

image


icons/north-star.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
colebemis: `fill="none"` is not needed here.

```suggestion
<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
```

keywords.json
```diff
@@ -136,6 +136,7 @@
   "mortar-board": ["education", "learn", "teach"],
   "mute": ["quiet", "sound", "audio", "turn", "off"],
   "no-newline": ["return"],
+  "north-star": ["star", "snowflake", "asterisk"]
```
colebemis: This will probably fix the failing build.

```suggestion
  "north-star": ["star", "snowflake", "asterisk"],
```

icons/north-star.svg
ashygee: d'oh! thank you! 🙏🏽

PR #388, opened 2020-03-14
Implement 24px icons

colebemis:
## Problem

The current Octicons don't scale up well. When rendered at larger sizes, they look a little out of place:

image

In Octicons v2, to address this issue, we designed 24px icons (in addition to 16px icons) to handle cases where we need larger icons:

<p align="center">
  Frame 1289
</p>

**However, our Ruby gems and React components now need to choose the right icon to render based on the size passed in.**

## Solution

Here's our plan to update the packages to handle multiple sizes per icon. This PR implements the parts outlined in red:

image

### Ruby

The only change to the public API of the Ruby gem is the structure of the objects in `OcticonsV2::OCTICONS_SYMBOLS`:

```diff
OcticonsV2::OCTICONS_SYMBOLS["x"]
{
   "name" => "x",
  "keywords" => [],
- "width" => 16,
- "height" => 16,
- "path" => "<path ... />",
+ "heights": {
+   "16" => {
+     "width" => 16,
+     "path" => "<path ... />",
+   },
  }
}
```

The rest of the API remains the same. We now use the `height` option passed into the `OcticonV2` constructor to determine the best SVG to render. Example: 

```ruby
OcticonsV2::OcticonV2.new("x")
{
  symbol: "x",
  keywords: [],
  width: 16,
  height: 16,
  options: {
    width: 16,
    height: 16,
    viewBox: '0 0 16 16',
    ...
  },
  path: "...path for 16px icon..."
}

OcticonsV2::OcticonV2.new("x", {height: 32})
{
  symbol: "x",
  keywords: [],
  width: 24,
  height: 24,
  options: {
    width: 32,
    height: 32,
    viewBox: '0 0 24 24',
    ...
  },
  path: "...path for 24px icon..."
}
```

### Rails and Jekyll

The source code for the Rails helper and Jekyll plugin was not changed. However, because of the updates to the Ruby gem, they will now automatically  render the best SVG based on the height passed in. 

### Website

The Octicons website now displays all the icon sizes:

| Home page | Icon page |
| --- | --- |
| localhost_8000_plus-circle-24_ | localhost_8000_plus-circle-24_ (1) |

### JavaScript

Updating the JavaScript and React packages may require significant changes to their public API. I'd like to implement those changes in a separate PR so they don't block the updates to the Ruby packages and the website. In other words, I'm proposing that we merge this PR into the `v2` branch even if the `node` and `react` checks aren't passing.


lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
```diff
@@ -84,5 +81,25 @@ def calculate_width(height)
     def calculate_height(width)
       (width.to_i * @height) / @width
     end
+
+    def get_octicon(symbol, options)
```
jonrohan: There's probably a newer cooler style of doing this, but you should guard against `options` build `nil` so line 87 doesn't fail when it is

```suggestion
    def get_octicon(symbol, options = {})
```

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
```diff
@@ -84,5 +81,25 @@ def calculate_width(height)
...
+          height: octicon_height,
+          path: octicon["heights"][octicon_height.to_s]["path"]
+        }
+      end
+    end
+
+    def closest_octicon_height(octicon_heights, height)
+      return octicon_heights.reduce(octicon_heights[0].to_i) {
+        |acc, octicon_height| (octicon_height.to_i <= height.to_i ? octicon_height.to_i : acc)
+      }
```
jonrohan: I would make this cleaner to read

```suggestion
      return octicon_heights.reduce(octicon_heights[0].to_i) do |acc, octicon_height| 
        octicon_height.to_i <= height.to_i ? octicon_height.to_i : acc
      end
```

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
```diff
@@ -84,5 +81,25 @@ def calculate_width(height)
...
+          keywords: octicon["keywords"],
+          width: octicon["heights"][octicon_height.to_s]["width"].to_i,
+          height: octicon_height,
+          path: octicon["heights"][octicon_height.to_s]["path"]
+        }
+      end
+    end
+
+    def closest_octicon_height(octicon_heights, height)
+      return octicon_heights.reduce(octicon_heights[0].to_i) do |acc, octicon_height|
```
colebemis:
```suggestion
      return octicon_heights.reduce(octicon_heights[0].to_i) do |acc, octicon_height|
```

lib/octicons_gem/test/octicons_test.rb
```diff
@@ -4,9 +4,8 @@
   it "loads all icons on initialization" do
     refute_equal 0, OcticonsV2::OCTICON_SYMBOLS.length
     x_icon = OcticonsV2::OCTICON_SYMBOLS["x"]
+    assert x_icon["name"]
     assert x_icon["keywords"]
-    assert x_icon["path"]
-    assert x_icon["height"]
-    assert x_icon["width"]
+    assert x_icon["heights"]
```
joelhawksley: Might we add a deep assertion for the content of heights?

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
```diff
@@ -5,14 +5,11 @@ class OcticonV2
...
-      if octicon = OcticonsV2::OCTICON_SYMBOLS[@symbol]
-
-        @path = octicon["path"]
-        @width = octicon["width"].to_i
-        @height = octicon["height"].to_i
-
-        @keywords = octicon["keywords"]
-
+      if octicon = get_octicon(@symbol, options)
+        @path = octicon[:path]
```
joelhawksley: Is there a reason we're switching to symbol keys? It might not matter in this case, but switching key formats has bitten us before, and I'd rather minimize the surface area of this change.

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
```diff
@@ -84,5 +81,25 @@ def calculate_width(height)
     def calculate_height(width)
       (width.to_i * @height) / @width
     end
+
+    def get_octicon(symbol, options = {})
+      if octicon = OcticonsV2::OCTICON_SYMBOLS[symbol]
+        height = options[:height] || options[:width] || 16
```
joelhawksley: What do you think about deprecating the `width` option?

joelhawksley:
```suggestion
        height = options[:height] || options[:width] || DEFAULT_HEIGHT
```

Generally I think it's best to not have magic numbers. Might we assign a constant for `16`?

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
```diff
@@ -84,5 +81,25 @@ def calculate_width(height)
...
+        octicon_height = closest_octicon_height(octicon["heights"].keys, height)
```
joelhawksley: Does every icon have a 24px version? If so, it could probably simplify this logic, which may run thousands of time on a single page 😬

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
colebemis: Makes sense. Where should `DEFAULT_HEIGHT` be defined?

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
colebemis: Right now, we can't assume that every icon has a 24px version. If it leads to a significant performance improvement, I'm open to enforcing that every icon has a 16px and 24px version. How would that simplify this logic?

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
joelhawksley: In this class, I think.

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
joelhawksley: This is a little rough, but maybe something as simple as:

```suggestion
        octicon_height = height > 16 ? 32 : 16
```

Using constants of course 😉

colebemis: Hmm, I'm hesitant about requiring all icons to have a 16px and 24px version for a couple reasons:

- Logos don't really need 16px and 24px versions since we can scale the logo svgs to any size without them looking weird.  Requiring that each logo has a 16px and 24px version (e.g. `mark-github-16.svg` and `mark-github-24.svg`) that are identical seems cumbersome.

- Hard coding the sizes that each icon must have would also make adding smaller or larger versions of icons in the future much more difficult. For example, adding a smaller sized icon (say 12px) would require us design a complete 12px set. This would make it hard for us to ship things in an "incrementally correct" way.

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
colebemis: We're now indexing into the hash table returned by the `get_octicon` function instead of the JSON object which had string keys. I can change the `get_octicon` function to return a hash with string keys if you think that's better.

lib/octicons_gem/test/octicons_test.rb
colebemis: Is there a good way to make deep assertions on `heights` without making assumptions about what keys it has?

colebemis: The type definition for `heights` would look something like this:

```ts
heights: {
  [key: string]: {
    width: number
    path: string
  }
}
```

lib/octicons_gem/lib/octicons_v2/octicon_v2.rb
colebemis: Is there a way we can measure the performance impact of the `closest_octicon_height` function?

joelhawksley: @colebemis those are fair points. I think you're right to push us towards benchmarking.

https://github.com/evanphx/benchmark-ips is what I use. I'd be happy to 🍐 on this ❤️

colebemis:
@joelhawksley Yeah, the CI is failing for the JavaScript and React packages because they haven’t been updated. Updating those packages may require significant changes to their public API. I'd like to make those changes in a separate PR so they don't block the updates to the Ruby packages and the website. In other words, I'm proposing that we merge this PR into the v2 branch even though the node and react checks aren't passing. How does that sound to you?

joelhawksley:
@colebemis how much work do you think it would take to make them pass? Could we not update them to just use the 16px icons for now, keeping our `v2` branch green?

colebemis:
> @colebemis how much work do you think it would take to make them pass? Could we not update them to just use the 16px icons for now, keeping our `v2` branch green?

Done ✅

lib/octicons_node/index.js
```diff
@@ -44,20 +53,22 @@ for (const key of Object.keys(data)) {
...
+  data[key].width = data[key].heights[DEFAULT_HEIGHT].width
+  data[key].height = parseInt(DEFAULT_HEIGHT)
 
   // Set all the default options
   data[key].options = {
     version: '1.1',
     width: data[key].width,
     height: data[key].height,
-    viewBox: `0 0 ${data[key].width} ${data[key].height}`,
+    viewBox: `0 0 ${data[key].heights[DEFAULT_HEIGHT].width} ${DEFAULT_HEIGHT}`,
```
colebemis:
```suggestion
    viewBox: `0 0 ${data[key].width} ${data[key.height}`,
```

Issue #393, opened 2020-04-03
RFC: Changing the Octicons React API to support 24px icons


colebemis:
As part of the [Octicons refresh](https://github.com/github/design-systems/issues/711) project, we designed a set of 24px icons (in addition to 16px icons). With the addition of this 24px grid size, we no longer have a 1-to-1 mapping of icon name to SVG file. The Octicons React components now need to choose which SVG to render based on the size passed in:

| Before                                                                                                         | After                                                                                                         |
| -------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- |
| ![image](https://user-images.githubusercontent.com/4608155/78410106-492bcf80-75c0-11ea-8af3-c7416b3ed2dc.png) | ![image](https://user-images.githubusercontent.com/4608155/78410127-56e15500-75c0-11ea-9ce7-0d9298b1c95b.png) |

The current public API of our Octicons React package makes implementing this change tricky. I'm proposing a simpler public API that would make this change easier to implement. Specifically, I'm purposing that we remove the `Octicon` "shell" component so each icon would be rendered with a single React component:

```diff
- <Octicon icon={Alert} />
+ <Alert />
```
I explain my reasoning for this API change below 👇 

## What is the public API right now?

We currently have 5+ ways to render an octicon in React:

1. `icon` prop (most common)

   ```jsx
   <Octicon icon={Alert} />
   ```

2. React `children`

   ```jsx
   <Octicon><Alert /></Octicon>
   ```

3. `getIconByName()` function

   ```jsx
   <Octicon icon={getIconByName('alert')} />
   ```

4. `iconsByName` object

   ```jsx
   <Octicon icon={iconByName.alert} />
   ```

5. `StyledOcticon` component

   ```jsx
   <StyledOcticon icon={Alert} />
   <StyledOcticon><Alert /></StyledOcticon>
   <StyledOcticon icon={getIconByName('alert')} />
   <StyledOcticon icon={iconsByName.alert} />
   ```

## Do we have to change the public API?

Probably. Let's see why.

We'll use this code example to anchor our discussion:

```jsx
<Octicon icon={Alert} size={24} />
```

The above snippet will render an `<svg>` on the page. Something like this:

```html
<svg ...>
  <path d="...">
</svg>
```

`Octicon` is responsible for rendering the `<svg>` "shell":

```html
<svg ...>

</svg>
```

`Alert` is responsible for rendering the elements inside the `<svg>`:

```html
<path d="...">
[...]

PR #395, opened 2020-04-09
Update React package to support 24px icons

colebemis:
## Motivation

As part of the [Octicons refresh](https://github.com/github/design-systems/issues/711) project, we designed a set of 24px icons—as well as 16px icons—to accommodate interfaces that need larger icons. With the addition of this 24px grid size, we no longer have a 1-to-1 mapping of icon name to SVG file. The Octicons React components now need to choose which SVG to render based on the size passed in:

| Before                                                                                                    | After                                                                                                        |
| -------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- |
| image | image |

This PR implements the "After" diagram. Here's how...

## Implementation

Our current React API for rendering an octicon looks like this:

```jsx
<Octicon icon={Alert} size={24} />
```

The above snippet will render an `<svg>` on the page:

```html
<svg viewBox="..." width="..." height="...">
  <path d="...">
</svg>
```

The `Octicon` component is responsible for rendering the `<svg>` "shell":

```html
<svg viewBox="..." width="..." height="...">

</svg>
```

The `Alert` component is responsible for rendering the elements inside the `<svg>`:

```html
<path d="...">
```

However, since there is no longer a 1-to-1 mapping of icon name to SVG file, `Alert` now needs access to the `size` prop passed to `Octicon` in order to decide whether to render `alert-16.svg` or `alert-24.svg`. And `Octicon` needs to know whether `Alert` chose to render `alert-16.svg` or `alert-24.svg` in order to compute the `viewBox` attribute for the `<svg>` element. 

To avoid creating a complex two-way data flow between `Octicon` and `Alert`, I've pushed all the rendering logic into the `Alert` component. `Alert`—and all the icon components—are now responsible for rendering the entire `<svg>`. `Octicon` just passes props to its icon component without rendering any additional elements, like so:

```jsx
function Octicon({icon: Icon, size}) => {
  return <Icon size={size} />
}
```

## Release notes

* Icon components (e.g. `AlertIcon`, `ArrowRightIcon`, etc.) now accept `size`, `ariaLabel`, `verticalAlign`, and `className` props and can be used on their own. No need to pass them to the `Octicon` component.
  ```jsx
  // Both ways work now:
  <Octicon icon={AlertIcon} size={24} />
  <AlertIcon size={24} />
  ```

* Icon components will now choose the best SVG icon to render based on the `size` passed in.

### BREAKING CHANGES 💥

* All icon component names now include `Icon` at the end (e.g. `Alert` → `AlertIcon`).

* `Octicon` no longer accepts `width` or `height` props. Use the `size` prop instead. In cases where the width and height of an icon are not equal (e.g. logos), the height will be set to the value of the `size` prop and the `width` will be scaled proportionally.

* Setting `verticalAlign="top"` on the `Octicon` component or any icon component will now apply a `vertical-align: top;` style to the `<svg>`. Previously, we were translating "top" to "text-top." So setting `verticalAlign="top"` would apply a `vertical-align: text-top;` style to the `<svg>`. I'm not sure why we did that. If you want a vertical alignment of "text-top," set the `verticalAlign` prop to `"text-top"`.

* Custom icon components passed to the `Octicon` component now need to render the entire `<svg>`, not just the `<path>`.

```diff
function CirclesIcon() {
  return (
-   <React.Fragment>
+   <svg viewBox="0 0 30 10" width="30" height="10">
      <circle r={5} cx={5} cy={5}/>
      <circle r={5} cx={15} cy={5}/>
      <circle r={5} cx={25} cy={5}/>
-   </React.Fragment>
+   </svg>
  )
}

- CirclesIcon.size = [30, 10]
```

## Next steps

The `Octicon` component now seems unnecessary and makes the API bloated.  Should we deprecate it?

If we deprecate the `Octicon` component, we would need to provide an alternative to the `StyledOcticon` component, which exposes system props for styling octicons. We could create a separate package called, say, `@primer/octicons-react-styled` that wraps every icon component with styled-systems props:

```jsx
import {Alert} from '@primer/octicons-react-styled'

<Alert color="gray.5" mr={2} />
```

I'd love to hear other people's thoughts on this 💭 

Closes https://github.com/primer/octicons/issues/393

---

Thank you to @dmarcey for early feedback on this implementation 🙏

### DIFF
--- a/lib/octicons_react/script/build.js
+++ b/lib/octicons_react/script/build.js
@@ -41,57 +41,9 @@ function writeIcons(file) {
   const count = icons.length
   const code = `${GENERATED_HEADER}
 import React from 'react'
+import getSvgProps from '../get-svg-props'
 
 ${icons.map(({code}) => code).join('\n')}
-
-function closestNaturalHeight(naturalHeights, height) {
-  return naturalHeights
-    .map(naturalHeight => parseInt(naturalHeight, 10))
-    .reduce(
-      (acc, naturalHeight) => (naturalHeight <= height ? naturalHeight : acc),
-      naturalHeights[0]
-    )
-}
-
-const sizeMap = {
-  small: 16,
-  medium: 32,
-  large: 64
-}
-
-function getSvgProps({
-  ariaLabel,
-  className,
-  size,
-  verticalAlign,
-  svgDataByHeight
-}) {
-  const height = sizeMap[size] || size
-  const naturalHeight = closestNaturalHeight(
-    Object.keys(svgDataByHeight),
-    height
-  )
-  const naturalWidth = svgDataByHeight[naturalHeight].width
-  const width = height * (naturalWidth / naturalHeight)
-  const path = svgDataByHeight[naturalHeight].path
-
-  return {
-    'aria-hidden': ariaLabel ? 'false' : 'true',
-    'aria-label': ariaLabel,
-    role: 'img',
-    className,
-    viewBox: \`0 0 $\{naturalWidth} $\{naturalHeight}\`,
-    width,
-    height,
-    fill: 'currentColor',
-    style: {
-      display: 'inline-block',
-      userSelect: 'none',
-      verticalAlign
-    },
-    dangerouslySetInnerHTML: { __html: path }
-  }
-}
 
 const iconsByName = {
   ${icons.map(({key, name}) => `'${key}': ${name}`).join(',\n  ')}
--- a/lib/octicons_react/script/build.js
+++ b/lib/octicons_react/script/build.js
@@ -122,8 +122,8 @@ type Size = 'small' | 'medium' | 'large'
 interface IconProps {
   ariaLabel?: string
   className?: string
-  size: number | Size
-  verticalAlign: 'middle' | 'text-bottom' | 'text-top' | 'top' | 'unset'
+  size?: number | Size
+  verticalAlign?: 'middle' | 'text-bottom' | 'text-top' | 'top' | 'unset'
 }
 
 type Icon = React.FC<IconProps>
--- a/lib/octicons_react/script/build.js
+++ b/lib/octicons_react/script/build.js
@@ -9,37 +9,29 @@ const typesFile = join(srcDir, 'icons.d.ts')
 
 const GENERATED_HEADER = '/* THIS FILE IS GENERATED. DO NOT EDIT IT. */'
 
-const DEFAULT_HEIGHT = '16'
-
-function CamelCase(str) {
+function pascalCase(str) {
   return str.replace(/(^|-)([a-z])/g, (_, __, c) => c.toUpperCase())
 }
 
-const octiconNames = [...Object.entries(octicons)]
-
-const icons = octiconNames
-  // Temporarily filter out icons without 16px SVGs until we update the library
-  // to handle mupltiple SVGs per icon.
-  .filter(([, octicon]) => DEFAULT_HEIGHT in octicon.heights)
+const icons = Object.entries(octicons)
   .map(([key, octicon]) => {
-    const name = CamelCase(key)
-    const width = octicon.heights[DEFAULT_HEIGHT].width
-    const height = parseInt(DEFAULT_HEIGHT)
-    const path = octicon.heights[DEFAULT_HEIGHT].path
-    // convert attributes like fill-rule into JSX equivalents, e.g. fillRule
-    const svg = path.replace(/([a-z]+)-([a-z]+)=/g, (_, a, b) => `${a}${CamelCase(b)}=`)
-    const type = `Icon<${width}, ${height}>`
-    const code = `function ${name}() {
-  return <React.Fragment>${svg}</React.Fragment>
+    const name = pascalCase(key)
+    const code = `function ${name}(props) {
+  const svgDataByHeight = ${JSON.stringify(octicon.heights)}
+  return <svg {...getSvgProps({...props, svgDataByHeight})} />
 }
-${name}.size = [${width}, ${height}]
+
+${name}.defaultProps = {
+  className: 'octicon',
+  size: 16,
+  verticalAlign: 'text-bottom'
+}
 `
 
     return {
       key,
       name,
       octicon,
-      type,
       code
     }
   })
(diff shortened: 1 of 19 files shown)


lib/octicons_react/script/build.js
```diff
@@ -52,6 +44,55 @@ import React from 'react'
...
+    height,
+    fill: 'currentColor',
+    style: {
+      display: 'inline-block',
+      userSelect: 'none',
+      verticalAlign
+    },
+    dangerouslySetInnerHTML: { __html: path }
+  }
+}
```
colebemis, line 94: This is a lot of static code in a template string. I'll pull it out into a file.

colebemis:
https://github.com/primer/octicons/runs/569898522#step:5:173

The `ts-tests` started failing and I'm not sure why. It looks like TypeScript is failing to type check a type definition file in `node_modules` but adding `exclude: ['node_modules']` in the `tsconfig.json` didn't fix it. Any ideas on how to fix the type errors, @dmarcey?

dmarcey:
> https://github.com/primer/octicons/runs/569898522#step:5:173
> 
> The `ts-tests` started failing and I'm not sure why. It looks like TypeScript is failing to type check a type definition file in `node_modules` but adding `exclude: ['node_modules']` in the `tsconfig.json` didn't fix it. Any ideas on how to fix the type errors, @dmarcey?

Hmmm, I cloned this repo locally and ran `npm run ts-test` and didn't get any errors. Does it work locally for you too? Is it just an error in the CI run?

dmarcey:
@colebemis I think I figured out the issue. The combination of bumping the `typescript` version in the package.json and adding `skipLibCheck` in the tsconfig.json seemed to fix the issue.

https://github.com/primer/octicons/pull/396

Feel free to merge this PR into yours or just make a single commit with the changes if you don't want all of my commits where I was trying to figure out what was wrong 😄 

I also "broke" the types to make sure that the build would still fail in this case: https://github.com/primer/octicons/pull/396/checks?check_run_id=574324407#step:5:173 and it did.

lib/octicons_react/script/build.js
```diff
@@ -76,15 +117,21 @@ function writeTypes(file) {
...
-  W extends number = number,
-  H extends number = number
-> = React.SFC<{}> & { size: [W, H] };
+type Size = 'small' | 'medium' | 'large'
+
+interface IconProps {
+  ariaLabel?: string
+  className?: string
+  size: number | Size
+  verticalAlign: 'middle' | 'text-bottom' | 'text-top' | 'top' | 'unset'
```
colebemis, line 126:
```suggestion
  size?: number | Size
  verticalAlign?: 'middle' | 'text-bottom' | 'text-top' | 'top' | 'unset'
```

emplums:
Hey @colebemis this makes sense to me! The only thing I'm worried about is the naming of the components themselves - could we prepend each component name with `Icon` so `Alert` is `AlertIcon`? That way we don't run into naming clashes with Primer Components (for instance, I could see us having an `Alert` component for alert UI at some point)

colebemis:
> could we prepend each component name with Icon so Alert is AlertIcon?

@emplums ✅ Done

colebemis:
There's still one big question we have to answer: **How should consumers be able to apply system props to octicons?**

Here are a few options I can think of:


###  Option 1: Keep `StyledOcticon`

```jsx
import {StyledOcticon} from '@primer/components'
import {AlertIcon} from '@primer/octicons-react'
<StyledOcticon icon={AlertIcon} color="red.6" mr={2} />
```

Pros:
* Simple to implement
* Avoids a breaking change in Primer Components

Cons:
* More typing/clunky API

### Option 2: Add system props to icon components in `@primer/octicons-react`

```jsx
import {AlertIcon} from '@primer/octicons-react'
<AlertIcon color="red.6" mr={2} />
```

Pros:
* No need to wrap the icon in a separate component just to apply styles

Cons:
* Forces consumers to use `styled-system` and `styled-components` under the hood

### Option 3: Create a separate package that wraps all the icon components with system props

```jsx
import {AlertIcon} from '@primer/octicons-react-styled'
<AlertIcon color="red.6" mr={2} />
```

Pros:
* No need to wrap the icon in a separate component just to apply styles
* People can still use `@primer/octicons-react` if they don't want to use `styled-system` and `styled-components`

Cons:
* People have to know to install `@primer/octicons-react-styled` instead of `@primer/octicons-react` if they want to use system props
* We have to keep the `@primer/octicons-react` and `@primer/octicons-react-styled` packages synced up, which could be tricky to implement in this monorepo.

---

I'd love to hear people's thoughts on this. Which option should we choose? Are there any pros or cons that I'm missing? Are there better options I haven't considered?

cc @broccolini @emplums @dmarcey

dmarcey:
From an API standpoint for a consumer, I like Options 2 and 3 the best. Wrapping with `<StyledOcticon />` feels clunky and inconsistent with other components (e.g. we don't have a `<StyledTextInput />`). 

For deciding between 2 and 3, I think we need to know more about how people use the package.

Do we know whether or not we have consumers of the `@primer/octicons-react` package that do not also consume `@primer/components`? If we do, or we want to support that scenario, then I would vote for Option 3.

If we're saying that `@primer/octicons-react` is really only meant to be used in conjunction with `@primer/components`, then I think it's fine for it to take the dependency on `styled-system` and expose the system props directly on the different icons.

Another question which might help lead us down the right path. We've discussed adding the `sx` prop to `@primer/components`. https://github.com/primer/components/projects/3#card-33491189
If we do, would we expect Octicons in react to support that prop as well? If so, how would we go about adding that support?

colebemis:
> There's still one big question we have to answer: How should consumers be able to apply system props to octicons? https://github.com/primer/octicons/pull/395#issuecomment-611767640

We discussed this in our DSUI review meeting today. The consensus was to move forward with Option 3 but the separate package should be thought of as an extension of Primer Components instead of an extension of the Octicons package.  This subtle change means that the package would probably be called something like `@primer/components-octicons` instead of `@primer/octicons-react-styled` and the code for the package would live outside the `primer/octicons` repo.

lib/octicons_react/src/get-svg-props.js
```diff
@@ -0,0 +1,36 @@
+const sizeMap = {
+  small: 16,
+  medium: 32,
+  large: 64
+}
+
+export default function getSvgProps({ariaLabel, className, size, verticalAlign, svgDataByHeight}) {
```
dmarcey, line 7: Should this prop be `aria-label` instead of `ariaLabel` . While we're making breaking changes it seems like this could be another spot to clean up and be more consistent with other components in Primer/components.

colebemis: Yeah., I agree that `aria-label` is the correct prop name to be consistent with the rest of React. I was a little worried about making that breaking change because it could break aria-labels without people noticing. However, now is probably a good time to rip off the bandaid and correct this API since we're already making other breaking changes. Okay, you convinced me. I'll change it 😸

colebemis: I'll make that change in a follow up PR though since it's not directly related to 24px icons.

Issue #404, opened 2020-04-21
Request to add support for tree-shaking in @primer/octicons-v2-react


dmarcey:
After consuming `@primer/octicons-v2-react` in the memex package, I noticed that the bundle size jumped about 130 kb (gzipped / minified). I thought that perhaps we didn't have tree-shaking working properly in our build system, so I played around with it a little bit, but couldn't quite get anything working.

So, I decided to start with a clean repo (based on `create-react-app`), and see if I could get it working there. When I was unable to, I decided to pull in a different icon library - `react-feather` to see if I could get tree-shaking working with it. I was able to with no extra configuration.

Here are the repro steps and the bundle size differences I encountered.

1. `git clone git@github.com:dmarcey/tree-shaking-test.git` (https://github.com/dmarcey/tree-shaking-test)
2. `cd tree-shaking-test`
3. `yarn install`
4. `yarn build` Bundle size: `39.39 KB build/static/js/2.a430f49c.chunk.js`
5. Uncomment lines 3 and 26 (`import` and usage of `Camera`).
6. `yarn build` Bundle size: `40.17 KB (+802 B)  build/static/js/2.b6d6f8f9.chunk.js`
7. Recomment lines 3 and 26
8. `yarn build`
9. Uncomment lines 4 and 27 (`import` and usage of `RepoIcon`)
10. `yarn build` Bundle size: `168.39 KB (+129 KB)  build/static/js/2.6febd43a.chunk.js`

This quick test shows us that including a single `react-feather` icon in your bundle adds about 800 bytes, whereas adding a single `@primer/octicons-v2-react` icon to your bundle adds 129 KB - so tree-shaking is not happening.

I stumbled across #335 by chance - where it looks like there was an attempt made to have each icon in its own file - which looks to be the same approach taken by `react-feather`, which are then all exported individual at a top-level `index.js`.
https://github.com/feathericons/react-feather/tree/master/src/icons 

Let me know if any more info is needed - we'd love to help get this resolved before octicons v2 goes live!

cc/ @colebemis @emplums

PR #405, opened 2020-04-22
Enable tree-shaking in React package

colebemis:
### Problem

Using a single icon from `@primer/octicons-v2-react` adds about 130kb to the bundle size of an application—so tree-shaking is not happening:

```
File sizes after gzip:

  169.18 KB (+129.79 KB)  build/static/js/2.fc93e4a5.chunk.js
  771 B                   build/static/js/runtime-main.f8c5b4be.js
  653 B (+7 B)            build/static/js/main.ea532662.chunk.js
  547 B                   build/static/css/main.5f361e03.chunk.css
```

@dmarcey wrote a detailed description of the problem in this issue: https://github.com/primer/octicons/issues/404

### Solution

Adding `"sideEffects": false` to the `package.json` and removing the `iconsByName` and `getIconByName` exports seemed to fix the tree-shaking issue. I tested these changes in @dmarcey's [tree-shaking-test](https://github.com/dmarcey/tree-shaking-test) repo and found that using one icon from `@primer/octicons-v2-react` only added 1.69kb to the bundle (verses 130kb before):

```
File sizes after gzip:

  41.08 KB (+1.69 KB)  build/static/js/2.220ad890.chunk.js
  771 B                build/static/js/runtime-main.f8c5b4be.js
  653 B (+7 B)         build/static/js/main.ea532662.chunk.js
  547 B                build/static/css/main.5f361e03.chunk.css
```

### Risks

Removing `iconsByName` and `getIconByName` from the React package would be a breaking change. Consumers would no longer be able to render a list of all the icons or dynamically get an icon component by name. However, @smockle noticed that there is only a single code match for `iconsByName` in the entire GitHub org—and it’s from the deprecated package `@githubprimer/octicons-react`: https://github.com/search?q=org%3Agithub+iconsByName&type=Code. So maybe this breaking change wouldn't be disruptive.

Closes https://github.com/primer/octicons/issues/404


dmarcey:
I can do the same later tonight for Memex. Do you have a package version handy that I can test with?

colebemis:
> Do you have a package version handy that I can test with?

`@primer/octicons-v2-react@0.0.0-be2ea3b`

PR #370, opened 2020-02-18
Octicons v2 👀

colebemis:
## What is this branch?

This `v2` branch is where we'll do all the code updates for the [Octicons v2](https://github.com/github/design-systems/issues/711) project. Keeping a long-lived feature branch like this allows us to update and publish new Octicons packages under different names (i.e. `octicons` → `octicons-v2`).  Publishing new Octicons packages will allow us to have both new and old Octicons packages installed on dotcom simultaneously. Once we're ready to remove the Octicons feature flag from dotcom, we can merge this branch into `master`. Before merging into master, we'll need to remember to revert the package names (i.e. `octicons-v2` → `octicons`).

🔍 [All PRs](https://github.com/primer/octicons/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Amerged+base%3Av2) merged into `v2` (so far).

## Packages

| Type | Current package | New package |
| --- | --- | --- |
| gem | [`octicons`](https://rubygems.org/gems/octicons) | [`octicons_v2`](https://rubygems.org/gems/octicons_v2) |
| helper | [`octicons_helper`](https://rubygems.org/gems/octicons_helper) | [`octicons_v2_helper`](https://rubygems.org/gems/octicons_v2_helper) |
| jekyll | [`jekyll-octicons`](https://rubygems.org/gems/jekyll-octicons) | [`jekyll-octicons_v2`](https://rubygems.org/gems/jekyll-octicons_v2) |
| node | [`@primer/octicons`](https://www.npmjs.com/package/@primer/octicons) | [`@primer/octicons-v2`](https://www.npmjs.com/package/@primer/octicons-v2) |
| react | [`@primer/octicons-react`](https://www.npmjs.com/package/@primer/octicons-react) | [`@primer/octicons-v2-react`](https://www.npmjs.com/package/@primer/octicons-v2-react) |

## How to use this branch

1. Make a PR and change the base branch to `v2`
    - You might wanna branch off from `v2` if your changes are related.
2. When the PR is approved, merge it
3. Update `octicons_v2` packages in dotcom

## Helpful links

- [Octicons refresh project details](https://github.com/github/design-systems/issues/711)
- [Octicons implementation plan](https://github.com/github/design-systems/issues/754)

PR #417, opened 2020-04-28
Create @primer/styled-octicons package

colebemis:
## Problem

We want to be able to style Octicons with [system props](https://primer.style/components/system-props) like we style other Primer Components. The components in `@primer/octicons-react`, however, don't accept system props. So we created a `StyledOcticon` component in Primer Components as a workaround. But the API for using `StyledOcticon` is clunky:

```
<AlertIcon /> // how you use an icon if you don't need system props
<StyledOcticon icon={AlertIcon} mr={2} /> // to use system props, you need to wrap the icon in StyledOcticon
```

## Solution

What if we could pass system props directly to icon components?

```diff
- <StyledOcticon icon={AlertIcon} mr={2} />
+ <AlertIcon mr={2} />
```

This PR makes that possible. It adds a `@primer/styled-octicons` package that wraps the components in `@primer/octicons-react` with [`COMMON`](https://primer.style/components/system-props) system props.

```jsx
import {AlertIcon} from '@primer/styled-octicons'

function Example () {
  return <AlertIcon mr={2} color="red.6" />
}
```

📝 [Preview the docs](https://octicons-git-styled-octicons.primer.now.sh/octicons/packages/styled-system)

You can test out the package by installing:

```
npm install @primer/styled-octicons@canary
```

After we merge this PR, we'll be able to deprecate `StyledOcticon` in Primer Components.

## Questions

* Should icon components accept the `display` prop?
  * I assume the answer is "yes" because `display` is part of the [`COMMON`](https://primer.style/components/system-props#system-prop-categories) system props in Primer Components.
  * If the answer is "yes," we'll have to update the `@primer/octicons-react` because it uses inlines styles to [set `display: inline-block`](https://github.com/primer/octicons/blob/v2/lib/octicons_react/src/get-svg-props.js#L23-L27), which overrides the `display` prop.  But how will we apply styles in `@primer/octicons-react` without using inline styles or styled components?
* Should icon components accept the `sx` prop?
* Are we happy with the name `@primer/styled-octicons` for this package? 
* What should the peer dependencies of this package be? I've set `react` and `styled-components` as peer dependencies. But should `styled-system` also be a peer dependency? 

## TODO

- [x] Write PR description
- [x] Write docs
- [x] Fix tree shaking
  - I tested the canary version of this package in @dmarcey's [tree-shaking-test](https://github.com/dmarcey/tree-shaking-test) repo and found that tree shaking was not working. I need to figure out tree shaking works for `@primer/octicons-react` but not for `@primer/styled-octicons`.
- [x] Add `sx` prop

---

Part of https://github.com/primer/components/issues/511


emplums:
> Should icon components accept the display prop?
Honestly I don't think this is a hard requirement. If Octicons are always meant to be displayed inline-block we can leave it that way.

> Should icon components accept the sx prop?
yes this would be nice to have.  though if we wanted to have a default theme for these icons we'd have to have a dep on @primer/components which we might not want to do

> Are we happy with the name @primer/styled-octicons for this package?
yes 🚢 it!

> What should the peer dependencies of this package be? I've set react and styled-components as peer dependencies. But should styled-system also be a peer dependency?
I think just react and styled-components should be fine but not sure about styled-system

emplums:
Ok just looked through the code and it got me thinking more about the theme. Should we assume that folks are using this library inside of a `ThemeProvider` somewhere that's hooked up to the Primer Component's theme? Should we supply a backup theme just in case like we do in Primer Components?

colebemis:
@emplums I think styled-system has a default spacing scale. So `<AlertIcon mr={2} />` will work even outside of a ThemeProvider. I'm okay with requiring a ThemeProvider to use colors. I think that's a better option than adding `@primer/components` as a dependency in order to set a default theme.

dmarcey:
@colebemis I can confirm I saw the same results as you with respect to tree-shaking. I played around with the package directly in my `node_modules` folder and I think I got something working when I moved the individual icons into their own files and then just exported them all from the root.

From this comment (https://github.com/webpack/webpack/issues/9337#issuecomment-524966201) it seems like maybe that's the best way to handle things anyway? I'm guessing since we're now executing a function call each icon (to compose it with system props), as opposed to just defining in the functions like in the `octicons-v2-react` package, the optimizer doesn't know what to omit.

However, it can use more of a "side-effects" approach which each icon is in its own module, perhaps?

Feel free to ping me if you want to talk through some of this synchronously and I can walk you through what I tried.

colebemis:
🎉 Bundling each icon individually fixed the tree-shaking issue! Thanks for your help, @dmarcey.

lib/octicons_styled/script/build.js
```diff
@@ -0,0 +1,59 @@
...
+
+interface IconProps extends StyledSystem.ColorProps, StyledSystem.SpaceProps {
+  'aria-label'?: string
+  className?: string
+  size?: number | Size
+  verticalAlign?: 'middle' | 'text-bottom' | 'text-top' | 'top' | 'unset'
+  sx?: SystemStyleObject
+}
+
+type Icon = React.FC<IconProps>
```
colebemis, line 30: Would it be helpful if exported these types? @dmarcey

lib/octicons_styled/script/copy.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+cp src/__generated__/index.d.ts dist/index.d.ts
+cp src/__generated__/index.js dist/index.esm.js
```
colebemis, line 4:
```suggestion
cp src/__generated__/index.js dist/index.esm.js

```

colebemis:
@dmarcey @emplums This is ready for final review 😸

docs/content/packages/styled-system.mdx
```diff
@@ -0,0 +1,44 @@
...
+import {AlertIcon} from '@primer/styled-octicons'
+
+export () => (
+  <AlertIcon color="red.6" mr={2} />
+)
+```
+
+## System props
+
+All icon components in `@primer/styled-octicons` get `COMMON` system props. Read our [System Props](https://primer.style/components/system-props) documentation for a full list of available props.
```
dmarcey, line 31: Should we explicitly call out support for the `sx` prop here?

lib/octicons_styled/README.md
```diff
@@ -0,0 +1,5 @@
+# @primer/octicons-react
+
+[![npm version](https://img.shields.io/npm/v/@primer/octicons-react.svg)](https://www.npmjs.org/package/@primer/octicons-react)
```
dmarcey, line 3: Do these need to be updated?

lib/octicons_styled/script/build.js
dmarcey: Yes! I think it would be nice to export both `IconProps` and `Icon`.

docs/content/packages/styled-system.mdx
colebemis: Yes. Forgot to add that. Nice catch :)

lib/octicons_styled/README.md
colebemis: Yep. Thank you!

docs/content/packages/styled-system.mdx
```diff
@@ -0,0 +1,44 @@
...
+---
+
+[![npm version](https://img.shields.io/npm/v/@primer/styled-octicons/canary.svg)](https://www.npmjs.org/package/@primer/styled-octicons)
+
+The `@primer/styled-octicons` package wraps icon components from [`@primer/octicons-v2-react`](/packages/react) with [system props](https://primer.style/components/system-props), making them easier to style in projects that use [styled system](https://styled-system.com/)—like [Primer Components](https://primer.style/components).
+
+## Install
+
+```shell
+npm install @primer/styled-octicons@canary
```
emplums, line 14: Why do we recommend folks to install the canary?

colebemis: Until the [`v2` branch](https://github.com/primer/octicons/pull/370) is merged, there won't actually be a `@latest` version of the package. When the `v2` branch is merged, we'll update the docs to remove `@canary`.

emplums: 🙌

Issue #240, opened 2018-09-05
Fill with color icon


diegohdez90:
Provide user add fill color to icon.

shawnbot:
Hi there! You can pass a `fill` attribute via the Ruby and Jekyll helpers, and a `fill` prop to the React component. Do you have a more specific question, or have you run into any issues passing `fill` to a specific library?

diegohdez90:
I was trying add `fill` property in my Octicon React component with this:

```
<Octicon
  icon={GitPullRequest}
  fill='green'
/>
```

also with:

```
<Octicon
   icon={GitPullRequest}
   style={
   {
       fill: 'green'
   }}
/>
```

and none display me Octicon grren
What is the correct use of `fill` property?

Thanks

shawnbot:
Hey, you're right! We're not passing along any additional props in the Octicon component. That's my bad. I'm labeling this as an issue and will get to it ASAP. Thank you for letting us know.

diegohdez90:
There is an advance. ????

shawnbot:
Not yet, no! Thanks for your patience, everyone. 🙇

diegohdez90:
No problem. 🤓

rutsky:
As a workaround you can set fill color from parent element:

```jsx
<span style={{color: "red"}}><Octicon><Check /></Octicon></span>
```

broesmel:
Hello! I think this issue still exists in current versions of primer. Any updates?

PR #418, opened 2020-04-29
React: fill is not a style property, it's a svg attribute.

macno:
Added a default to currentColor but it's possibile to overwrite it via 'fill' prop

Updates react test

It should fix also #240

### DIFF
--- a/lib/octicons_react/src/index.js
+++ b/lib/octicons_react/src/index.js
@@ -15,7 +15,7 @@ const alignMap = {
 const defaultSize = [16, 16]
 
 export default function Octicon(props) {
-  const {ariaLabel, children, className, height, icon: Icon, size, verticalAlign, width} = props
+  const {ariaLabel, children, className, height, icon: Icon, size, verticalAlign, width, fill} = props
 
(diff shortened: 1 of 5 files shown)


lib/octicons_react/src/index.js
```diff
@@ -26,6 +26,7 @@ export default function Octicon(props) {
     'aria-label': ariaLabel,
     className,
     height,
+    fill: fill || 'currentColor',
```
colebemis, line 29: Could we give `fill` a default value with `defaultProps` instead of using the `||` operator?

macno:
@colebemis no, thank you for the project :)

lib/octicons_react/src/index.js
macno: Done

macno:
@colebemis Hi, I did all the changes suggested, do they look good to you?

colebemis:
@macno Sorry for the delay. We're working on releasing a new major version of Octicons that includes all the new icons you see on github.com. I'll probably bundle this change with that release.

colebemis:
Closing in favor of #551

PR #422, opened 2020-05-06
Add more migration info to CHANGELOG

colebemis:
### Problem

In Octicons v2, some v1 icons have been renamed and there are still some icons that were in v1 that we don't have in v2 yet. We need to document the mapping between v1 to v2 icons to help people who want to upgrade from v1.

### Solution

I originally planned to create a separate migration guide. But as I was writing it, I realized that I was repeating a lot of information that was already in the CHANGELOG. So I decided to add the table mapping v1 to v2 icons to the CHANGELOG instead of making a separate document for it.

 I think we'll eventually want to make the CHANGELOG more discoverable from https://primer.style/octicons.


CHANGELOG.md
```diff
@@ -1,5 +1,62 @@
...
+| `ruby`                   |                           | Need this icon? Let us know by [opening an issue](https://github.com/primer/octicons/issues/new?assignees=&labels=icon+request&template=icon-request.md&title=%5BIcon+request%5D) |
+| `saved`                  | `bookmark`                |                                                                                                                                                                                   |
+| `squirrel`               |                           | Need this icon? Let us know by [opening an issue](https://github.com/primer/octicons/issues/new?assignees=&labels=icon+request&template=icon-request.md&title=%5BIcon+request%5D) |
+| `text-size`              | `heading` or `typography` |                                                                                                                                                                                   |
+| `unsaved`                | `bookmark-slash`          |                                                                                                                                                                                   |
```
ashygee, line 53: Do we want to mention `triangle-...` has been renamed to `caret-...`?

colebemis: Good catch! I'll add that. Thank you :)

PR #425, opened 2020-05-14
updating for center alignment with other icons

ashygee:
Fixing to be center aligned when in a vertical layout with other icons

before | after
:--: | :--:
old | new


colebemis:
Hmm, this new alignment looks off when centered in a circle:

image

https://octicons-nr82wk142.now.sh/octicons/link-external-16#ui-examples

I think it might be better to stick with what we had. What do you think?

ashygee:
> Hmm, this new alignment looks off when centered in a circle:
> 
> image
> 
> https://octicons-nr82wk142.now.sh/octicons/link-external-16#ui-examples
> 
> I think it might be better to stick with what we had. What do you think?

@colebemis Ah I see what you're saying! Hmmm... let's just close this and keep it as is. I will recommend the use of the `link` icon instead in the other UI we noticed earlier.

PR #427, opened 2020-05-21
change url for new octicons-v2

ashygee:
updated link to point to url within primer.style domain.


colebemis:
`primer.style/octicons-v2` points to the icons that in the `primer/octicons-v2` which are no longer up-to-date. Here's a diagram that shows the current state of things:

image

Our two options moving forward are to update the alias `primer.style/octicons-v2` to point to the most up-to-date v2 icons or revert this PR. I think updating the alias is probably the better option.

ashygee:
@colebemis I think that's the better option too. Are you able to do this?

PR #440, opened 2020-06-08
Add paintbrush icon

colebemis:
image


colebemis:
Should `paintbrush` and `pencil` face the same direction?
image

PR #448, opened 2020-06-10
Add some v1 icons to v2 branch 

colebemis:
In order to remove the `octicons_v2` feature flag on github.com, we need to have a complete mapping of v1 icons to v2 icons. We are currently missing v2 replacements for browser, bug, device-camera, repo-clone, and repo-pull. This pull copies over those icons from v1 so we can remove the `octicons_v2` feature flag in github.com.

image

This is a temporary fix. We'll need to design v2 versions for these eventually.


FloEdelmann:
Can you please update the release notes for v10 accordingly?

PR #451, opened 2020-06-11
docs: Fix the example for the deprecated Octicon

suhaibmujahid:
The example of the deprecated usage of Octicon component has wrong suffix for the icon name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,7 @@
 - The `Octicon` component is deprecated. Use icon components on their own instead:
 
   ```diff
-  - <Octicon icon={AlertIcon} />
+  - <Octicon icon={Alert} />
   + <AlertIcon />
   ```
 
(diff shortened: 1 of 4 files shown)


colebemis:
Hi, @suhaibmujahid! Thanks for the PR. These changes are kind of confusing so thanks for bearing with us 😅  

There were two changes to the Octicons React API that are relevant to this PR:
1. All icon component names now include `Icon` at the end (e.g. `Alert` → `AlertIcon`).
2. The `Octicon` component is deprecated.

These changes aren't connected to each other. So if you were to use the Octicon component now (which you shouldn't), you do it like this: `<Octicon icon={AlertIcon} />` (not `<Octicon icon={Alert} />`). So the documentation is technically correct, although I agree it's confusing. I'm open to ways of clarifying this in the docs.

Hopefully this made sense :)

PR #457, opened 2020-06-25
Save caret-down-16.svg

ashygee:
This adjusts the `caret-down-16` icon to be one pixel lower.

This image shows a side by side comparison of the updated dropdown caret. left is old with caret centered in container, right is with the caret moved 1px down


ashygee:
I just realized we've named this icon `triangle-down` instead of `caret-down`. I will update the Figma library and rename this icon as to not mess with our source in production.

cc @edokoa

PR #458, opened 2020-06-26
Save octoface-24.svg, octoface-16.svg

ashygee:
This updates the octoface to fit with the rest of the new octicons styling

Comparison between new and old octoface. Left: old octoface with only eyes, Right: new octoface with full face, eyes and smiling mouth


edokoa:
@ashygee @ohjoycelau Approved to have the new version up, but what do you think about making the mouth corners in the 16px version subtly rounded?  They'd virtually look the same at 16px but I think it looks a little bit better and more consistent with 24px in High density displays or when scaling up for whatever reason.

https://www.figma.com/file/fDx33rXo60Hn41hiX17754/octoface-icon-exploration?node-id=101%3A117

image

image


We can patch it if you want.

ashygee:
@edokoa that works . We don’t need to patch. Just make a commit to this branch.

edokoa:
@ashygee Done!

PR #465, opened 2020-07-13
Rename icon assets

BenJetson:
Greetings, Octicons maintainers!

I was writing a small library that uses Octicons today and noticed a couple of small problems with your file names.

Firstly, I corrected this typo: `unverifed-24.svg` → `unverified-24.svg`.
Secondly, I renamed `file-symlink-24.svg` → `file-symlink-file-24.svg`.

The goal of the second rename is to make the names more consistent. The smaller asset was already named `file-symlink-file-16.svg` and the entry in `keywords.json` is for `file-symlink-file`.

Let me know if I need to update these names elsewhere in the repository. I did a global search and didn't see any references, so I think the rename is sufficient.

Cheers,
Ben


BenJetson:
Hi all, I have just updated this PR with changes for a couple of other icons that had inconsistent names.

`fire-24.svg` → `flame-24.svg`
`eye-slash-24.svg` → `eye-closed-24.svg`

Both of these now match the name in `keywords.json`.

However, I would note that `eye-slash` would probably be a better name for that icon, given that others like it use that convention (for example `bell-slash`, `bookmark-slash`, and `circle-slash`). I'd be interested in hearing your thoughts about this.

Best,
Ben

joelhawksley:
@colebemis this should be a major release, due to the breaking API changes 👍

PR #489, opened 2020-08-24
Updated release docs

ashygee:
Added updated release documentation.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -72,7 +72,7 @@ git checkout -b release-x.y.z
 
 In the context of Octicons, the version number is updated as follows:
 
-- Significant changes to the library or workflow, renaming or removing an icon would all be considered a **major update**. 
+- Breaking changes to the library or workflow, renaming or removing an icon would all be considered a **major update**. 
 - Adding a new icon would be considered a **minor update**.
 - Fixing an icon would be considered a **patch**.
 
@@ -125,6 +125,5 @@ When all of the checks have passed and the release PR has been approved, merge t
 
 
 🎉 Congratulations! The new release has been published.
-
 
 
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -70,22 +70,61 @@ Once submitted changes have been agreed upon, these instructions will guide main
 git checkout -b release-x.y.z
 ```
 
-In the context of Octicons, significant changes to the library or workflow, or removing an icon would be considered a major update. Adding a new icon would be considered a minor update. Fixing an icon would be considered a patch. Reach out in the #design-systems Slack channel if you're unsure.
+In the context of Octicons, the version number is updated as follows:
+
+- Significant changes to the library or workflow, renaming or removing an icon would all be considered a **major update**.
```
colebemis, line 75: I feel like it's easier to determine if a change is "breaking" rather than "significant." "Significant" seems a little vague to me.

```suggestion
- Breaking changes to the library or workflow, renaming or removing an icon would all be considered a **major update**. 
```

.github/CONTRIBUTING.md
```diff
@@ -70,22 +70,61 @@ Once submitted changes have been agreed upon, these instructions will guide main
...
-This will update `package.json` with the new version, then update all the `lib/*` packages with the same version. If that runs smoothly, it should commit the changed files.
+**Example:**
 
-### 4. Push your branch and open a pull request into `master`.
+![image](https://user-images.githubusercontent.com/10384315/91103190-c6171e80-e61f-11ea-8396-7138996cff30.png)
 
-A GitHub action will automatically publish a canary release of each Octicons package for each commit pushed to a branch. If the branch is prefixed with `release-` it will publish a release candidate.
 
-### 5. Merge feature branches into the release branch
+🎉 Congratulations! The new release has been published.
```
colebemis, line 127: Should we add a section at the end about updating Dotcom? I think we can add that when we walkthrough this process with @edokoa when he gets back.

ashygee: Let's discuss when @edokoa is back. I wasn't sure if we should add that information here.

PR #491, opened 2020-08-27
Cache Octicon SVGs for same icon/options combination

joelhawksley:
This PR upstreams an optimization we've been using
internally to speed up rendering the same octicon
multiple times.

Rendering octicons can be expensive when done often on a page. For
example, rendering the same "check" octicon in a large list of items can
add up on a page, so often we'll render it once and then pass in the
rendered HTML string into the list items.

Co-authored-by: Simon Taranto - @srt32


lib/octicons_helper/lib/octicons_helper/helper.rb
```diff
@@ -2,13 +2,23 @@
...
     return "" if symbol.nil?
 
-    icon = Octicons::Octicon.new(symbol, options)
-    content_tag(:svg, icon.path.html_safe, icon.options) # rubocop:disable Rails/OutputSafety
+    cache_key = [symbol, options].to_s
+
+    if octicons_helper_cache[cache_key]
+      octicons_helper_cache[cache_key]
+    else
+      icon = Octicons::Octicon.new(symbol, options)
```
jonrohan, line 17: Do you think we should include any of the raise conditions here when the symbol doesn't exist?

joelhawksley: @jonrohan Good idea! I think we should look into it in another PR.

PR #500, opened 2020-09-22
Update Rails and Jekyll package descriptions

colebemis:
Updates a few strange sounding sentences that @yaili pointed out.


yaili (approved):
🚢

PR #518, opened 2020-12-02
Save strikethrough-24.svg, strikethrough-16.svg

edokoa:


ashygee (approved):
This looks great! Thanks @edokoa 🚀 it!

PR #527, opened 2020-12-17
Fix Safari rendering of arrows

### LINKED ISSUE #485: Safari fold / unfold rendering issues
On a 1x screen in Safari, the arrows in fold icons appear cut off on one side, making these look like thumbs.

image
image

aaronshekey:
This PR closes #485 by removing fill-rules from the problematic arrow icons. This can be seen on 1x screens running Safari. It's also visible in quickview on macOS.

I also re-authored the `desktop-download-16` to assume a non-zero fill rule, which Safari expects. Seems other browsers do not have this assumption. I think the root cause is these icons were authored with non-zero fills, but then the SVG toggles them to be `evenodd`. Safari seems to be the only browser that cares about this discrepancy.

image


aaronshekey:
👋

colebemis (approved):

colebemis:
Thank you, @aaronshekey ❤️

PR #534, opened 2021-01-11
Multi select

Juliusschaeper:
Added 16 + 24px multi select icons.


colebemis:
Thanks for opening this PR, @Juliusschaeper! The different spacing in spacing between the caret and the dot looks a little off to me in the 24px version. I'm curious to see what it would look like if the spacing was equal.

image

Juliusschaeper:
@colebemis Good catch, thanks! I increased the spacing to 2.5px between stroke and chevron. 3px felt a bit too much because of the angled shape of the chevron.

colebemis (approved):
Nice work, @Juliusschaeper!

PR #543, opened 2021-02-04
Update bug icon

edokoa:
Updating bug icon as per https://github.com/github/design-systems/issues/998


colebemis:
@edokoa Looks like this PR contains 24 and 48 pixel icons instead of 16 and 24.

edokoa:
@colebemis I had the icons aligned and I pushed the wrong ones. It should be fixed now.

colebemis (approved):
🐞

Issue #539, opened 2021-01-20
[Icon request] Codespaces


mkwng:
Hello! We've been using `desktop-device` as a stand-in for Codespaces throughout the interface, and it's about time we got ourselves an official icon, don't you think? I don't think we have to overthink this, I just want to make sure we're not mis-applying icons in the future.

### What should the icon represent?

Some thoughts:
* We've been using the `desktop-device` icon, and it's mostly conveying the right thing
* Something that captures the idea of a workbench, a place you return to regularly to get work done
* Something that evokes the cloud or the virtual machine aspect of this would be an interesting alternative direction

To avoid:
* A direct connection to Visual Studio Code

### Where will the icon be used?

See https://github.com/features/codespaces

### What is the timeline?

Before end of Feb would give us a lot of breathing room

### Stretch goals?

This is absolutely **not** necessary, but we sometimes need a shorthand way of indicating the status of a Codespace. This can, of course, be conveyed outside of the icon, BUT if there's an elegant way for a Codespace icon to also capture the idea, that would be cool, no?
* Neutral
* Actively running
* Suspended
* Working/thinking... (building/rebuilding or deleting)
* Brand new

edokoa:
@mkwng We were working on different icons and decided on this one

![image](https://user-images.githubusercontent.com/694831/107244677-94a7d900-6a2e-11eb-8011-201c8e24e902.png)

![image](https://user-images.githubusercontent.com/694831/107244732-a2f5f500-6a2e-11eb-91a8-8421522ed467.png)


I created this PR #546

edokoa:
test comment

ashygee:
> @mkwng We were working on different icons and decided on this one

For clarity, @edokoa had several ideas and from those ideas the icon example shared above is what we had decided to test in #546 . The Codespaces icon is still a WIP and review conversation and status updates will take place in the PR. See our [Icon process ](https://github.com/primer/octicons/blob/master/.github/CONTRIBUTING.md#3-icon-design-review-and-communication)for more details.

edokoa:
We updated the PR with some icons and added the latest version of the Codespaces icon we were discussing in our last Octicons sync.

| Example |
|--------|
| ![UI example with small codespace icon](https://user-images.githubusercontent.com/10384315/108304278-89565b00-715c-11eb-845b-8ab6b8ba39af.png) |
[...]

PR #546, opened 2021-02-08
Save codespaces-16.svg

edokoa:
Codespaces icon for the request in #539


ashygee:
@mkwng Here's an idea that @edokoa has been working on for Codespaces. The [preview link](https://octicons-git-codespaces-16-v1.primer.now.sh/) can show you the details for the icon and here are a few screenshot using the Codespaces menu from the Code tab

image (11)
image (10)

edokoa:
I added this one to the PR as 'codespaces2'

image

mkwng:
Awesome! Below, you'll find screenshots of how this works in context. I think we _could_ run with codespace2, though if you're open to another round of revisions I have some feedback.


image  codespace 
- It feels a bit muddy in context, too much happening
- Looks like a document
- Feedback from the team: "The icon doesn’t clearly convey the idea codespaces to me… I think it could be improved by incorporating some kind of cloud imagery to the icon, for example"
- Not in love with how it looks in the context of a button (see below)


image codespace2
- Personally feel this one is closer, but it's very similar to the code icon if we're worried about brand-ability
- Still not in love with how it works within the button
- It patterns weird on longer lists, making me reconsider using it in that context
- Looks a bit funny at different scales, but i guess this wouldn't be a problem if we only use it at 16pt

To emphasize the stretch goals, it would be great if we had variations for when codespaces are live/running, vs suspended, but i also understand if that's beyond the scope of this exercise.

Thanks for your work so far! Lmk what you think, curious how you feel seeing the icons in context.


## codespace
| Description | Image  |
|---|----------|
| Code dropdown | image | 
| Code dropdown: Empty prompt | image |
| Button | image |
| Button Dropdown with icon in list | image |
| Conflict Dialog | image |
| Codespaces list | image |





## codespace2
| Description | Image  |
|---|----------|
| Code dropdown | image | 
| Code dropdown: Empty prompt | image |
| Button | image |
| Button Dropdown with icon in list | image |
| Conflict Dialog | image |
| Codespaces list | image |

edokoa:
test comment

edokoa:
test 2

edokoa:
I pushed `codespaces-3` to test.

image
image

We are going to discuss this today but I wanted to share it, there are some size issues I'd like to address, like why do the icons in the screenshots look 36x36 instead of 32x32, and how we could make this icon work with the diagonal lines, which affect the visual size.

ashygee:
Following up with today's octicons sync I've pushed `codespaces-v4` that @edokoa created. The context behind this version (and also v3 seen above) is that it represents two spaces that overlap to show their connectedness, much like codespaces with a local development environment. With the overlap two chevrons are created on the outer edges which is another connection to code. Examples in UI are seen below:

image

image

@mkwng while this metaphor is more abstract, it feels the best out of the options we've explored thus far as it has the connection to multiple spaces, code, and also feels unique enough to represent codespaces on its own.

We also feel this icon will not only be sustainable to represent Codespaces, but also scalable to offer us opportunities to explore showing when a codespace is running (on) vs suspended (off-ish) by playing with fills to represent state in the future.

edokoa:
I didn't update the PR but here are different filled versions of that icon:

https://www.figma.com/file/2rjsQp42uKvzPXzLZIqOeA/Octicons-v2-Source?node-id=5703%3A10192

image

ashygee:
@edokoa I got some good feedback from @ampinsk on this:

> I think if the code brackets were more clear I would have gotten that more

ashygee:
@edokoa I'm going to close this PR and create a new one for the new icon we are rolling with.

PR #550, opened 2021-02-23
Adding pull_request to the list of triggers only on main

adrianmg:
This should make the checks run whenever an outside collaborator tries to contribute to opening a PR. Right now it was only running when pushes were made.

cc @colebemis


colebemis:
❤️ Thank you, @adrianmg!

colebemis (approved):

PR #551, opened 2021-02-23
[React] Add fill prop

colebemis:
## Problem

Our octicon components don't provide a way to change the `fill` attribute of the underlying SVG.

## Solution

This PR adds a fill prop to all octicon components. @macno originally implemented this in #418 but merge conflicts made that PR difficult to merge.

Co-authored-by: @macno

### DIFF
--- a/docs/content/packages/react.mdx
+++ b/docs/content/packages/react.mdx
@@ -93,7 +93,7 @@ export default () => (
 ### Fill
 
 The `fill` prop takes a string value that can be used to set the color of the icon.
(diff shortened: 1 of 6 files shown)


jonrohan (commented):
Mostly want to know what others think about adding fill. Does this mean someone will be able to color the icon with fill and with color props?

colebemis:
@jonrohan Our components don't currently accept a `color` prop. Right now, the only ways to change the color of an icon with the React library is to style an icon with CSS or change the color of a parent element.

jonrohan (approved):

PR #565, opened 2021-03-24
Use more explicit dependencies for Rails helper

dbussink:
This doesn't require all of Rails, only ActionView & Railties. Reducing the dependencies here means that it enables folks to use only parts of Rails and to slim down their applications by removing unused components of Rails.


joelhawksley (approved):

dbussink:
@joelhawksley Looks like CI didn't run for some required builds, do they need to be explicitly triggered?

joelhawksley:
@dbussink I'm not sure. @colebemis mind having a look here?

dbussink:
> @dbussink I'm not sure. @colebemis mind having a look here?

@colebemis Any idea what is up here and if we can get this merged?

PR #569, opened 2021-03-31
Add sidebar expand and collapse icons

colebemis:
image


ashygee:
@colebemis is `collapse` and `expand` the correct descriptor for this? not `open`/`close`?

colebemis:
@ashygee I'm fine with either.

colebemis:
`expand`/`collapse` feel like they represent the action of "expanding" or "collapsing" which is how the icons will be used. `open`/`close` feel more like they could be states of the sidebar.

ashygee:
> `expand`/`collapse` feel like they represent the action of "expanding" or "collapsing" which is how the icons will be used. `open`/`close` feel more like they could be states of the sidebar.

This makes perfect sense. let's keep as is then! 🚀 it!

PR #592, opened 2021-04-12
Number icon

Juliusschaeper:
This PR modifies the current `number` icon with the new SVG and renames the old SVG to `hash`.


colebemis:
Nice work, @Juliusschaeper! The icons look good.

It looks like this PR replaces the old `#` icon which I don't think is what we want. We should decide what we want to call the `#` icon now or if we want to rename this icon.

ashygee:
@Juliusschaeper we need to replace the current `number-24` with the new `hash-24` name too.

ashygee (approved):
Looks good to me! 👍🏼

colebemis:
`number` icon looks great 👍 

The top-left and bottom-right steps on the `hash` icon look a little long to me though:
CleanShot 2021-05-12 at 09 25 58

PR #602, opened 2021-04-21
Creating CI workflow

jonrohan:
I wanted to separate out the build and the ci, so I can work toward a system that's more inline with [primer/css workflows](https://github.com/primer/css/tree/main/.github/workflows) Where there'll be changesets, ci, and releases separate.


colebemis (approved):
LGTM 👍

PR #614, opened 2021-05-13
Refresh issue icons

ashygee:
This adds refreshed issue icons

image


colebemis (approved):

laserlemon:
Out of curiosity, where did `issue-in-progress` go? We have a use case for it in our app. And `issue-reopened` seems to have been left out of the conversion from exclamations to dots. Was that intentional?

Andre601:
I have to ask, why `issue-closed` was accepted as it is right now?

That icon is basically the same as `check-circle`

I quickly made this comparison by downloading the 24px version of both icons from the site and importing them into Gimp at 500px size. Left one is `check-circle` and right one `issue-closed`
comparison

PR #617, opened 2021-05-13
Add row and column icons

ashygee:


colebemis (changes requested):
Looks like the dimensions of these icons are 16x17 and 24x25. They should be 16x16 and 24x24

colebemis (approved):

PR #646, opened 2021-08-07
Fix high CPU usage in docs

johanholmerin:
Noticed high CPU usage from the [icon page](https://primer.style/octicons/star-24) of the Octicon documentation. DevTools confirmed activity, see attached image and trace. Tracked it down to `IconPage` being constantly rerendered. Caused by passing an object created in the component as a dependency to `React.useEffect`, which uses shallow compare by default, causing the hook callback to be executed on every render. Since the callback calls `setPdf`, triggering a render, this causes an infinite loop.

Fixed by using `pageContext` as a dependency instead, which is what the local object is created from. Other solutions exist, e.g. only creating the PDF when the download button is clicked.

DevTools trace with high CPU

[DevTools Trace](https://github.com/primer/octicons/files/6949614/high-cpu-octicons.json.zip). Load in Performance tab of Chrome DevTools to view.

### DIFF
--- a/docs/src/templates/icon-page.js
+++ b/docs/src/templates/icon-page.js
@@ -31,7 +31,7 @@ export default function IconPage({pageContext}) {
     () => {
       getPdf(icon).then(blob => setPdf(blob))
     },
-    [icon]
+    [pageContext]
   )
 
   const [copied, setCopied] = React.useState(false)


colebemis (approved):
Thank you, @johanholmerin 💖

PR #655, opened 2021-08-15
Fix incorrect jekyll-octicons install instructions

misode:
Following the instructions I kept getting a `Unknown tag 'octicon'` error on build. Then I finally checked the [jekyll plugin documentation](https://jekyllrb.com/docs/plugins/installation/) and it noted that you have to list your plugins under `plugins` in `_config.yml`, not `gems`.


colebemis (approved):
Good catch. Thank you, @misode!

PR #665, opened 2021-09-22
Rename select-single-16.svg to single-select-16.svg

colebemis:
Rename `select-single-16` to `single-select-16` to be consistent with the 24px version (`single-select-24`).

Fixes https://github.com/github/primer/issues/173


ashygee (approved):
🚢

PR #667, opened 2021-09-28
Save iterations-16.svg and iterations-24.svg

edokoa:
Added iterations 16px for https://github.com/github/primer/issues/280


six7 (approved):
Awesome, thanks! 🎉

PR #671, opened 2021-10-04
Update and rename docs/content/index.mdx to ドキュメント/コンテンツ/index.mdx

utowa9544:


ashygee:
Hi @utowa9544, thanks for the PR. We aren't able to change this using the Japanese characters but we will look into the link change on our end. I'm closing this for now.

PR #674, opened 2021-10-13
Changing rails helper example to view component

jonrohan:
I want to change the [Ruby](https://primer.style/octicons/alert-16#ruby) implementation example on the octicons pages to point to usage of the [Octicon View Component](https://primer.style/view-components/components/octicon). Our goal in dotcom is to deprecate use of the helper and probably deprecate the package afterwards.

### DIFF
--- a/docs/src/templates/icon-page.js
+++ b/docs/src/templates/icon-page.js
@@ -104,9 +104,9 @@ export default function IconPage({pageContext}) {
           <H2>Usage</H2>
           <Paragraph>
             You can use the{' '}
-            <Link to="https://primer.style/view-components/components/octicon">
-              Octicon View Component
-            </Link>
+            <a href="https://primer.style/view-components/components/octicon">
+              View Component
+            </a>
             ,{' '}
             <Link as={GatsbyLink} to="/packages/jekyll">
               Jekyll helper


benkoshy:
@jonrohan Hi Jon

In order to use the Octicons without the rails helper - I presume that the `require "octicons"` must be placed in an initializer file somewhere (?) so that  `Primer::OcticonComponent.new(:check)` is available in rails views (erb)? Right now, I am getting a `uninitialized constant #<Class:0x0000558aabfe5408>::Primer` error, was wondering how that could be fixed: a pointer would be appreciated. chrs,  Ben

jonrohan:
> @jonrohan Hi Jon
> 
> 
> 
> In order to use the Octicons without the rails helper - I presume that the `require "octicons"` must be placed in an initializer file somewhere (?) so that  `Primer::OcticonComponent.new(:check)` is available in rails views (erb)? Right now, I am getting a `uninitialized constant #<Class:0x0000558aabfe5408>::Primer` error, was wondering how that could be fixed: a pointer would be appreciated. also i'm happy to work on a PR if required.  chrs,  Ben
> 
> 
> 
> 

You should be able to follow the directions for installation here. https://primer.style/view-components/#installation

PR #708, opened 2021-12-17
Added first batch of feed icons

Juliusschaeper:
Based on this request https://github.com/github/primer/issues/471 we’re adding the first batch of feed specific icons 

Bildschirmfoto 2021-12-15 um 18 17 29


ashygee (approved):
🚢

PR #754, opened 2022-02-23
Save trophy-16.svg, trophy-24.svg

edokoa:
This is the Achievements icon requested in https://github.com/github/primer/issues/629

I called it `Trophy` but I'm not sure @Juliusschaeper didn't name it `Achievement` already.


ashygee (approved):
🏆 🚢

PR #756, opened 2022-02-24
Cloud icons

Juliusschaeper:
Added `cloud` and `cloud-offline` icons based on this request https://github.com/github/primer/issues/411
service-outage


colebemis:
I'm not the biggest fan of the "overlay" in the clouds. It makes it feel a bit cartoon-y IMO. Have you tried versions without this overlap? Happy to join a working session to discuss more :)
CleanShot 2022-02-24 at 10 42 00@2x

Juliusschaeper:
@colebemis We [explored](https://www.figma.com/file/2rjsQp42uKvzPXzLZIqOeA/?node-id=7195%3A59122) several directions without the overlay but ultimately decided that it makes the cloud more recognizable as such.

Juliusschaeper:
Here is a comparison between overlay and closed strokes. @ashygee suggests a more reduced stroke overlay (#2)
service-outage-iteration

edokoa:
As I said, I still feel like the overlay strokes make them not fit into the set. I don't dislike the left one without strokes so much. I kind of recall the one I didn't like was more symmetric? We can discuss it again at the Sync next Tuesday.

Juliusschaeper:
Updated cloud shape to a single outline service-outage

ashygee (changes requested):
@Juliusschaeper something funky is happening with the SVGs and they're appearing only as solid black square when I view the files. @colebemis can you help?

preview of cloud icons as black squares

colebemis:
The `x` in `cloud-offline` looks a little small when rendered at 16x16. Have you tried using a slash? Similar to the `bell-slash` icon
CleanShot 2022-03-03 at 09 05 15

Juliusschaeper:
@colebemis @ashygee @edokoa I’m not sure why we haven’t considered this, but I do like how it makes it more consistent and readable. Thoughts?

A-cloud
B-cloud
A-preview
B-preview

edokoa:
I agree with the slash.

Watching it in mobile it seems that the slash is optically misaligned to the right (at 16px).

Juliusschaeper:
Pushed an update with improved optical alignment of the `\`
B-cloud - B

icons/cloud-24.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path d="M0 0h24v24H0z"/><path fill-rule="evenodd" d="M3.103 10.107c0-4.244 3.445-7.607 7.733-7.607 3.19 0 5.912 1.858 7.099 4.563a.634.634 0 01.01.022l.001.006C21.348 7.345 24 10.095 24 13.536 24 17.148 21.076 20 17.431 20H5.017C2.23 20 0 17.83 0 15.06a4.9 4.9 0 013.112-4.581 8.024 8.024 0 01-.009-.372zM10.836 4c-3.485 0-6.233 2.717-6.233 6.107 0 .284.022.602.052.756a.75.75 0 01-.552.869c-1.52.385-2.603 1.712-2.603 3.328 0 1.917 1.532 3.44 3.517 3.44h12.414c2.843 0 5.069-2.206 5.069-4.964 0-2.759-2.226-4.965-5.069-4.965a.75.75 0 01-.696-.47l-.179-.446C15.606 5.5 13.424 4 10.836 4z"/></svg>
```
ashygee, line 1: Black square on this one

ashygee (changes requested):

.changeset/cold-swans-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@primer/octicons": patch
```
ashygee, line 2:
```suggestion
"@primer/octicons": minor
```

ashygee: This should be a minor update because we are adding new icons. A patch would be used for something like when we update a current icon's vectors but it is still named the same.

ashygee (changes requested):

ashygee (approved):
🚀

PR #763, opened 2022-03-03
Save person-remove-16.svg

edokoa:
Adds the icon to replace the trash icon when removing users from repository settings 
* https://github.com/github/primer/issues/736


ashygee (approved):
🚀

ashygee:
This icon is no longer going to be used.

PR #785, opened 2022-05-11
Add Copilot icons

ashygee:
16 and 24 pixel copilot icons that look like little airplane pilots with goggles

This PR adds the official octicons for Copilot.

This closes https://github.com/github/primer/issues/902

PR #788, opened 2022-05-11
Revert "Add Copilot icons"

jonrohan:
Reverts primer/octicons#785


colebemis (approved):
Why are we reverting this?

PR #801, opened 2022-06-29
Save paperclip-16.svg, paperclip-24.svg

gavinmn:
Adds the new `attachment` icon using the paperclip metaphor.

attachment icons


gavinmn:
> Looks great! Just curious, did you explore a rotated paperclip icon? Similar to this:
> 
> CleanShot 2022-06-29 at 14 23 27@2x
> 
> I wonder how the straight vs rotated versions compare when placed in context.

That's a good idea. I hadn't looked at that yet so we can hold of on merging this and I'll take a second look to see if it works better.

gavinmn:
Revisited this icon exploring out the tilted variation: 
attachment

I think this version is more distinct and makes better use of the horizontal space.

ashygee (approved):

PR #803, opened 2022-07-13
fix table icon

ashygee:
This PR fixes the table icon to remove overhang that caused blurring pixels.

overhang icon example


edokoa (approved):
👍

PR #811, opened 2022-07-28
Save command-palette-24.svg, command-palette-16.svg

ashygee:
Adds icon to represent command palette vs search

command palette icons represented as command prompt with > and _


edokoa (approved):
\>_<

PR #837, opened 2022-09-15
[Experiment] replace usage of `dangerouslySetInnerHTML` in `octicons-react`

joshblack:
## Background

https://github.com/primer/octicons/issues/796

I was looking into `dangerouslySetInnerHTML` usage in Primer and wanted to try an experiment with `@primer/octicons-react` that replaced the usage of `dangerouslySetInnerHTML` with JSX elements. This is being opened as a draft to gauge if the approach aligns with any existing efforts in this space 👀 

## What's changed

At a high-level, this PR changes a component from being defined as:

```jsx
function AccessibilityIcon(props) {
  var svgDataByHeight = { "16": { "width": 16, "path": "<path fill-rule=\"evenodd\" d=\"M9.923 5.302a3 3 0 10-3.847 0A2.713 2.713 0 005.9 5.5H2A.75.75 0 002 7h3.3l-.578 5.163-.362 2.997a.75.75 0 101.49.18L6.132 13h3.736l.282 2.34a.75.75 0 101.49-.18l-.362-2.997L10.7 7H14a.75.75 0 000-1.5h-3.899a2.697 2.697 0 00-.178-.198zM9.5 3a1.5 1.5 0 11-3 0 1.5 1.5 0 013 0zm-.3 4.073l.495 4.427h-3.39l.496-4.427a1.207 1.207 0 012.398 0z\"></path>" } };
  return React.createElement('svg', getSvgProps(_extends({}, props, { svgDataByHeight: svgDataByHeight })));
}

AccessibilityIcon.defaultProps = {
  className: 'octicon octicon-accessibility',
  size: 16,
  verticalAlign: 'text-bottom'
};
```

To the following:

```jsx
const AccessibilityIcon = createIconComponent("AccessibilityIcon", "octicon octicon-accessibility", {
  "16": {
    "width": 16,
    "path": /*#__PURE__*/React.createElement("path", {
      fillRule: "evenodd",
[...]

PR #838, opened 2022-09-16
refactor(octicons_react): update package deps and tests

joshblack:
This PR introduces some prep work before https://github.com/primer/octicons/pull/837 moves forward. In particular, it adds in some Public API and tree-shaking tests to protect against any regressions if an export no longer exists or the bundle can no longer be tree-shaken.

**What's changed**

- Add PublicAPI.test.js which snapshots the exports from the entrypoint
- Add tree-shaking.test.js which verifies that unused exports can be tree-shaken with no bundle impact
- Update babel deps
- Update rollup deps
- Update Next.js preview
  - Update Next.js to v12
  - Update preview to use the new icon format
- Update octicon test to no longer emit `console.warn` for the deprecated `Octicon` export


lib/octicons_react/__tests__/PublicAPI.test.js
```diff
@@ -0,0 +1,7 @@
+const Octicons = require('@primer/octicons-react')
+
+describe('@primer/octicons-react', () => {
+  it('should not update exports without a semver change', () => {
+    expect(Object.keys(Octicons).sort()).toMatchSnapshot()
```
colebemis, line 5: ❤️

lib/octicons_react/__tests__/tree-shaking.test.js
```diff
@@ -0,0 +1,30 @@
...
+        throw new Error(message)
+      }
+    }
+  })
+  const {output} = await bundle.generate({
+    format: 'esm'
+  })
+
+  for (const {code} of output) {
+    expect(code.trim()).toBe('')
```
colebemis, line 28: If I'm reading this correctly, this test asserts that unused imports don't get bundled. Do we also need to check that importing and use a single icon doesn't bring all the icons with it?

colebemis (approved):

lib/octicons_react/__tests__/PublicAPI.test.js
```diff
@@ -0,0 +1,7 @@
+const Octicons = require('@primer/octicons-react')
```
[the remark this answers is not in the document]
joshblack: @colebemis honestly not really haha. I guess Public API felt like a proper noun? I'll update it over to `public-api.test.js` though 👍

lib/octicons_react/__tests__/tree-shaking.test.js
joshblack: @colebemis would be great to add, I'm just not sure what the assertion would look like 🤔 let me know if you have any ideas!

PR #848, opened 2022-09-30
Add filled alert icons

ashygee:
This adds fill versions for both the 16px and 24px alert icons

alert fill icons alongside of current alert icons


edokoa (approved):
I'll leave a comment in Slack / an issue about reviewing this kind of icon (with inverted symbols) later to consider making the line slightly thicker to counter the fill, but we will have to take into account the icon is also inverted in dark mode and other things...

* https://github.com/github/design-infrastructure/issues/2789

PR #863, opened 2022-11-03
ci: add workflow for updating package API snapshots

joshblack:
Add in a workflow for updating the package API snapshots for `octicons-react` without having to pull down the PR and run snapshot tests. 

This workflow does use the `pull_request_target` event but guards against misuse by requiring a label to be placed on the Pull Request in order to run. This technique comes from: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ and allows us to check that someone with write permission to the project has "approved" this workflow to run because the label is on the Pull Request.


.github/workflows/package_api.yml
```diff
@@ -0,0 +1,48 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request.head.repo.full_name == 'primer/octicons' && contains(github.event.pull_request.labels.*.name, 'Update Public API') }}
```
colebemis, line 16: Nitpick: How do you feel about calling the label `api changes approved`? (I don't think we use capital letters in any of our labels)

colebemis (approved):
Love this workflow! Just left a minor comment about the label naming

.github/workflows/package_api.yml
joshblack, line 16:
```suggestion
    if: ${{ github.event.pull_request.head.repo.full_name == 'primer/octicons' && contains(github.event.pull_request.labels.*.name, 'api changes approved') }}
```

colebemis (approved):
Ship it 🚀

PR #873, opened 2022-11-22
Add Goal icons

gavinmn:
Closes https://github.com/github/primer/issues/1445

image


colebemis:
Love it! 😍 

Have you tried using a 0.25px corner radius on the end of the arrow for consistency with other icons?

CleanShot 2022-11-22 at 14 34 41@2x

Here's the section about that in the design guidelines: https://primer.style/octicons/guidelines/design#corners

gavinmn:
@colebemis that's a great call! I had originally tried this but got a result that was much too round so I went with a 0 px radius. I just realized tho that it was because I was applying the rounding to the icon before flattening it. I just commited some new icons that should fix this inconsistency.

gavinmn:
@colebemis is this good to go?

colebemis (approved):
Ship it 🎯

Issue #866, opened 2022-11-07
[Bug] Some icons render incorrectly or throw warnings when used in iOS, macOS xcasset bundles


eliperkins:
### Describe the bug

I put the SVG assets from the v17.9.0 release of Octicons into an xcasset bundle, enabling "Preserve Vector Data", and [built a viewer for the Octicons within a macOS app](https://github.com/github/octiconsapp).

A number of the icons rendered odd paths, resulting in improper display of the icons.

#### List of affected icons

- `broadcast-16`
- `broadcast-24`
- `copilot-48`
- `copilot-96`
- `copilot-warning-16`
- `eye-closed-24`
- `file-badge-16`
- `gear-24`
- `history-16`
- `issue-draft-16`
- `issue-draft-24`
- `issue-reopened-16`
- `iterations-16`
- `iterations-24`
- `link-16`
- `link-24`
- `mention-16`
- `mention-24`
- `meter-16`
- `moon-16`
- `number-24`
- `paintbrush-16`
- `paperclip-16`
- `paste-16`
- `paste-16`
- `people-16`
- `people-24`
- `person-16`
- `person-24`
- `person-add-16`
- `person-add-24`
- `person-fill-16`
- `person-fill-24`
- `question-16`
- `rss-16`
- `rss-24`
- `squirrel-16`
- `squirrel-24`
- `tools-24`
- `unmute-16`
- `unmute-24`
- `unverified-16`
- `webhook-16`

### Steps to reproduce

For Hubbers:
1. Run [the Octicons.app from `github/octicons`](https://github.com/github/octiconsapp)

For non-Hubbers:
1. Create a new iOS or macOS app
2. Add any of the affected Octicons to the app as an SVG asset within an xcassets bundle
3. Render the Octicon within the app

### Expected behavior

Octicons render on iOS and macOS in the same manner that they do within web browsers, without the need for modifications.

### Screenshots
_Please add screenshots to help explain the problem._

| `gear-24` | `history-16` | `issue-draft-16` | `iterations-24` | `person-24` |
| --- | --- | --- | --- | --- |
[...]

PR #876, opened 2022-12-01
remove fill-rule="even-odd"

gavinmn:
Related to https://github.com/primer/octicons/issues/866

I was able to use https://github.com/googlefonts/picosvg to fix the `fill-rule` property of all of our Octicons. One thing to consider tho is that it looked like `picosvg` optimized strictly for filesize and the resulting icons have different point placements which makes the icons appear no different to the user, but theoretically would make editing an SVG from the library a bit clunkier. An example: 

| Pre-optimized icon | Fixed icon |
|--------|--------|
|CleanShot 2022-12-01 at 15 23 58@2x |CleanShot 2022-12-01 at 15 24 27@2x | 

Would love to hear some thoughts on these changes and if we want to consider adding something like this script to all icon pull requests to automate making sure we don't use `fill-rule="evenodd"`. An alternative discussed in https://github.com/primer/octicons/issues/866 was to add a CI test to catch `evenodd` which would require manual changing via the script or a Figma plugin before opening the pull request.

cc @lesliecdubs @eliperkins @tallys


colebemis:
This is great, @gavinmn! Could `picosvg` be a drop-in replacement for `svgo` in our `optimize.yml` workflow? https://github.com/primer/octicons/blob/main/.github/workflows/optimize.yml

We run that workflow on every push and commit the changes (if there are any).

gavinmn:
I don't know a ton about SVG optimization, but I don't _think_ `picosvg` is doing as good a job as `svgo`? See the difference in this diff — https://github.com/primer/octicons/pull/876/commits/1e0fdff1f0a4bf96fb03240b52f534bf70391b3f#diff-bbeeefbb309f5267472a458b4fe7e89a4f11ecba103fd5ccfc8616f823a199e5

Left is the `picosvg` output which is useful for removing `evenodd` but on the right, the `svgo` optimization on top of that looks much better. To be fair tho, I don't totally know my way around `picosvg` and there might be more options to get better results! Does `svgo` have any options to remove `evenodd` during optimization?

colebemis:
@gavinmn Ah, I see. We could run them both in that workflow: do the first pass with `picosvg`, then run `svgo`.

gavinmn:
I'll dig further into this utility and see if there's any way to avoid moving / adding vector points. I think otherwise if we are in agreement that having these slightly funky svg point layouts is not an issue we could add it to the workflow to avoid any `fill-rule` issues.

gavinmn:
From what I can tell, neither plugin would get us everything we want in isolation —`picosvg` doesn't result in great compression alone, and it doesn't look like any `svgo` plugin would remove / swap the error causing `fill-rule`.  Are there any technical tradeoffs that would need to be made to run both? We can fix this issue and then make sure we don't use `evenodd` in Figma with a CI test to check for it as an alternative if implementing `picosvg` would cause problems.

colebemis:
>  I think otherwise if we are in agreement that having these slightly funky svg point layouts is not an issue we could add it to the workflow to avoid any fill-rule issues.

I'm not worried about adding extra SVG points. These SVGs are the output of "outline stroke" in Figma, so they're already not the "source" file.

> Are there any technical tradeoffs that would need to be made to run both?

I don't see any downside besides that it will take longer to run, which I'm not concerned about.

Go ahead and add `picosvg` to the workflow 👍

gavinmn:
Going to work with Cole on getting this added as I'm not entirely sure how to add a python package to the workflow.

eliperkins:
Happy to mob on this with y'all too, should you need an extra set of eyes! I'm no Python expert, but can lend a hand if it'd be helpful ❤️

colebemis:
Gonna go ahead and close this in favor of @eliperkins's PR: https://github.com/primer/octicons/pull/880

PR #882, opened 2022-12-16
Use parameter defaults instead of defaultProps

manuelpuyol:
React is going to deprecate `defaultProps` in function components: https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-defaultprops-on-function-components

Instead of using `defaultProps`, we can rely on default parameters instead!


colebemis (approved):

manuelpuyol:
@colebemis any chance we could get this merged and released? 😄

PR #889, opened 2023-01-03
Update LICENSE to 2023

rezrah:
Fixes failing tests on `main` branch, by updating the License to 2023 instead of 2022. 

- Updated LICENSE
- Added changeset so that the updated license goes into the next release

### DIFF
--- a/.changeset/smooth-cameras-help.md
+++ b/.changeset/smooth-cameras-help.md
@@ -1,0 +1,5 @@
+---
+'@primer/octicons': minor
+---
+
+Updated LICENSE to 2023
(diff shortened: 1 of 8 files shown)


.changeset/smooth-cameras-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@primer/octicons': minor
```
colebemis, line 2: This probably doesn't need a changeset

colebemis (approved):
Happy New Year :🎊

.changeset/smooth-cameras-help.md
rezrah: Yeah probably, it's there to trigger a new release so we can replace the outdated one sooner rather than later.

rezrah: I'll remove it. It should have been a `patch` anyway 🤦

PR #894, opened 2023-01-11
[Bug] up/down chevron alignment

langermank:
Closes https://github.com/github/primer/issues/1673

This PR adjusts where the SVG sits within the frame to improve horizontal alignment between these related icons.

**Before**

image

**After**

image

**Concerns**

- Will this break how the icons behave individually? This might be considered an edge case where they appear side by side
- Should we take another approach to handle this at the component level?


gavinmn:
Going to add this to the Octicons agenda for next Wednesday.

tallys (approved):
Haven't heard any concerns about aligning the icon. If we can bundle this with addressing any visual regressions in buttons (if that was the original reason for the misalignment) let's do it.

langermank:
Here's a preview of how these changes will impact Primer React https://primer-3ab97043a8-13348165.drafts.github.io/storybook/?path=/story/components-iconbutton-features--chevron-test

gavinmn (approved):

PR #898, opened 2023-01-17
Delete package_api.yml

joshblack:
Remove the `package_api` workflow based on: https://github.com/github/primer/issues/1597


colebemis (approved):
😢

PR #914, opened 2023-01-30
adding correct naming conventions for Tracks and Tracked By

omerbensaadon:


gavinmn:
This would be a breaking change release that I'm not entirely sure how to navigate myself. @colebemis @tallys any thoughts or advice here?

omerbensaadon:
Bumping this one more time, I don't think anyone is using these icons if it brings you any solace

PR #924, opened 2023-02-28
Save shield-slash-24.svg

CameronFoxly:
Adding the 24px version of `shield-slash`, requested in [this issue](https://github.com/github/primer/issues/1676).


gavinmn (approved):
Looks good to me. I think this one is suffering from the "sliding top half" more than the 16px version, but we haven't found a good solution for addressing this issue quite yet.

colebemis (approved):

PR #925, opened 2023-03-01
Add `types` exports field for TypeScript ESM

kidonng:
This is [required](https://www.typescriptlang.org/docs/handbook/esm-node.html#packagejson-exports-imports-and-self-referencing) for TypeScript to work under `"module": "node16"`, see https://github.com//microsoft/TypeScript/issues/47792.


colebemis (approved):
Thanks @kidonng! ✨

PR #927, opened 2023-03-02
Save passkey-fill-24.svg, passkey-fill-16.svg

gavinmn:
Closes https://github.com/github/primer/issues/1672

Adds the Passkey icons

image


tallys (approved):

colebemis:
Should this be called `passkey-fill`?

gavinmn:
> Should this be called `passkey-fill`?

🤔 That's a good question. It's technically a filled icon, but it's not a filled icon with an outline icon counterpart since it's made specifically to follow the passkey spec. @tallys any opinions since this is kind of half a branded icon?

hagould:
We need "darkmode" versions of these as well for visibility

Ex:
Screen Shot 2023-03-03 at 2 15 44 PM

gavinmn:
We don't do different colors or modifications for Octicons in dark mode as far as I know, but once these are merged and released to the library they should be able to be used the same way you're currently using Octicons. Does that work or do you need to use the SVGs earlier?

tallys:
@gavinmn and @colebemis let's stick to our naming conventions and give it 'fill' title for consistency, even though there's not a counterpart. This is a bit of an edge case, but I think consistency in the library makes more sense

hagould:
> they should be able to be used the same way you're currently using Octicons

Ah thanks, I didn't know this. In that case this should be fine, thank you so much!

hagould (approved):

PR #928, opened 2023-03-03
Add discussion state Octicons

gavinmn:
Closes https://github.com/github/primer/issues/1895

image


gavinmn:
Hearing some feedback that when looking at the icons in context, I think I agree with — the state details are _really small_. I tried more variations to make this work and think this set may be better. 
image

antn:
I think those look great, @gavinmn! Let's go with those.

gavinmn:
> I think those look great, @gavinmn! Let's go with those.

Working on one more "reopen" state to add and then I'll get these out the door.

gavinmn:
Here's a new full proposal for the state icons, increasing the state accessory size by incorporating them into the icon and adding a `discussion-reopen` icon. 
image

gavinmn:
Shipping
image
We're discarding the `reopen` variant in favor of using no icon, or the `discussion-open` icon.

Issue #907, opened 2023-01-27
Accept additional props for labeling icons


joshblack:
We should make sure the React icon components accept additional props for labeling icons, including:

- [ ] `aria-labelledby`
- [ ] Support for adding a `<title>` tag

For the title tag, I think we have two options:

- A `title` prop that accepts either a `string` or `React.ReactNode` (in case the user wants to provide an `id` to the `title`
- Accept `children` as a prop and users can provide a title as children to the icon component

mattcosta7:
@joshblack I think we probably accept all of `React.SVGAttributes<SVGElement>` on the icon as well, to allow passing additional things like `aria-labelledby` if necessary, or even just `data-` attributes and the like

joshblack:
@mattcosta7 I'd totally be down for the `forwardRef` change 🔥 

I made https://github.com/primer/octicons/issues/910 to track this. Do you think it'd be something we should try and feature detect to ship it in a minor release or do you think it's worth trying to do a major release for the icons package and increase the deps to >=16.3?

mattcosta7:
> @mattcosta7 I'd totally be down for the `forwardRef` change 🔥 
> 
> I made https://github.com/primer/octicons/issues/910 to track this. Do you think it'd be something we should try and feature detect to ship it in a minor release or do you think it's worth trying to do a major release for the icons package and increase the deps to >=16.3?

I haven't tried to feature detect it, but no objections if there's a good path there!

pveierland:
Came here to file the same issue. It also seems sensible to set the value of `aria-label` to the `title` value if one is provided.

broccolinisoup:
👋🏻 @green6erry Are you actively working on this ticket? I see a linked PR but that was back in March. I have been working on some octicon changes https://github.com/primer/octicons/pull/943 and might have some capacity to pick this up if you are not planning to continue working on this? Totally cool if you are though! 🙌🏻  Just wanted to ask!

lesliecdubs:
@green6erry - ping to check out the draft PR and see if we can move it toward review. Thanks!

PR #932, opened 2023-03-15
907 accept additional props for labeling icons

green6erry:


lib/octicons_react/src/index.d.ts
```diff
@@ -1,14 +1,19 @@
+// eslint-disable-next-line import/no-namespace
 import * as React from 'react'
 
-import {Icon} from './__generated__/icons'
+import { Icon } from './__generated__/icons'
 
 type Size = 'small' | 'medium' | 'large'
 
 export interface OcticonProps {
```
joshblack, line 8: Would it makes sense to extend `React.SVGAttributes<SVGElement>` here based on: https://github.com/primer/octicons/issues/907#issuecomment-1406793154 to support additional props for the `<svg>` element?

lib/octicons_react/src/index.d.ts
```diff
@@ -1,14 +1,19 @@
...
   'aria-label'?: string
+  'aria-labelby'?: string
```
joshblack, line 10:
```suggestion
  'aria-labelledby'?: string
```

lib/octicons_react/src/index.d.ts
```diff
@@ -1,14 +1,19 @@
...
   tabIndex?: number
   children?: React.ReactElement<any>
   className?: string
+  title?: string | React.ReactElement<any>
+  ref?: React.Ref
```
joshblack, line 15:
```suggestion
```

I think we might need to exclude `ref` support unless we can find a way to safely use `React.forwardRef` (which seems tricky since this package supports v16)

lib/octicons_react/src/index.d.ts
```diff
@@ -1,14 +1,19 @@
...
 
 type Size = 'small' | 'medium' | 'large'
 
 export interface OcticonProps {
   'aria-label'?: string
+  'aria-labelby'?: string
   tabIndex?: number
   children?: React.ReactElement<any>
   className?: string
+  title?: string | React.ReactElement<any>
```
joshblack, line 14: Do you have a preference between `title` as a prop used for `<title>` or accepting `children` as a prop so people can supply it manually? e.g.

```tsx
<SomeIcon title="Some icon" />

// Versus
<SomeIcon>
  <title>Some icon</title>
</SomeIcon>
```

I assume people are okay with the former version but just wanted to ask.

joshblack (changes requested):
Looking great! Left some quick comments on this, curious what you think 👀

lib/octicons_react/src/index.d.ts
green6erry: Hey, I think since this is for icons, that title would be just a prop for accessibility reasons.

lib/octicons_react/src/index.d.ts
green6erry: Added the ability for it to be an svg 👍

lib/octicons_react/src/createIconComponent.js
```diff
@@ -12,24 +12,31 @@ export function createIconComponent(name, defaultClassName, getSVGData) {
 
   function Icon({
     'aria-label': ariaLabel,
+    'aria-labelledby': arialabelledby,
     tabIndex,
     className = defaultClassName,
     fill = 'currentColor',
     size = 16,
+    ref,
```
joshblack, line 20: I think `ref` should be covered in the latest updates from `main` with `React.forwardRef`: https://github.com/primer/octicons/blob/main/lib/octicons_react/src/createIconComponent.js#L23 once the branch is updated with latest 👀

lib/octicons_react/src/index.d.ts
```diff
@@ -1,16 +1,21 @@
...
   'aria-label'?: string
+  'aria-labelledby'?: string
   tabIndex?: number
   children?: React.ReactElement<any>
   className?: string
+  title?: string | React.ReactElement<any>
+  id?: string
   fill?: string
-  icon?: Icon
+  icon?: Icon | React.SVGAttributes<SVGElement>
```
joshblack, line 18: Just curious, what is the `React.SVGAttributes` bit used for here in `OcticonProps`?

lib/octicons_react/src/createIconComponent.js
```diff
@@ -12,24 +12,31 @@ export function createIconComponent(name, defaultClassName, getSVGData) {
...
+    id,
+    title,
     verticalAlign = 'text-bottom'
   }) {
     const height = sizeMap[size] || size
     const naturalHeight = closestNaturalHeight(heights, height)
     const naturalWidth = svgDataByHeight[naturalHeight].width
     const width = height * (naturalWidth / naturalHeight)
     const path = svgDataByHeight[naturalHeight].path
+    const svgProps = {...(id ? {id} : {}), ...(ref ? {ref} : {})}
```
joshblack, line 30: With the changes to `ref` I bet we could remove `svgProps` here and just set `id` explicitly on `<svg>` below

lib/octicons_react/README.md
```diff
@@ -62,6 +62,70 @@ export default () => (
 )
 ```
 
+### `aria-labelledby`
+
+You have the option of adding accessibility information to the icon with the
+[`aria-labelledby` attribute][aria-labelledby] via the `aria-labelledby` prop. Using aria-labelledby referencing the id values of the title element provides the accessible name.
```
joshblack, line 68: What is the link here intended to refer to? Specifically in: `[`aria-labelledby` attribute][aria-labelledby]`

lib/octicons_react/README.md
```diff
@@ -62,6 +62,70 @@ export default () => (
...
+
+```js
+// Example usage
+import {PlusIcon} from '@primer/octicons-react'
+
+export default () => (
+  <button>
+    <PlusIcon aria-labelledby="title" title="Add new item" /> New
```
joshblack, line 76: This example comes across confusing as we have the `button` and have three ways in which an accessible name can be computed:

- `aria-labelledby`
- `title`
- The text content `New`

Unfortunately I think `aria-labelledby` doesn't have a solid story when used within the context of a `button`. Is there a good alternate example that we could use for this scenario? Also happy to remove it as an explicit example if there isn't one handy.

lib/octicons_react/README.md
```diff
@@ -62,6 +62,70 @@ export default () => (
...
+You have the option of adding accessibility information to the icon with the
+[`title` attribute][title] via the `title` prop.
+
+```js
+// Example usage
+import {PlusIcon} from '@primer/octicons-react'
+
+export default () => (
+  <button>
+    <PlusIcon title="Add new item" /> New
```
joshblack, line 92: I think this example may lead to a problematic accessible name for the `button` where it would combine the title "Add new item" with "New". Is there a better example that we could reference for usage with `title`?

lib/octicons_react/README.md
```diff
@@ -62,6 +62,70 @@ export default () => (
...
+```js
+// Example usage
+import {PlusIcon} from '@primer/octicons-react'
+
+export default () => (
+  <button>
+    <PlusIcon ref="Add new item" /> New
+  </button>
+)
+```
```
joshblack, line 111: I think we can remove this section and just treat this as typical `React.forwardRef` guidance

lib/octicons_react/README.md
```diff
@@ -62,6 +62,70 @@ export default () => (
...
+```js
+// Example usage
+import {PlusIcon} from '@primer/octicons-react'
+
+export default () => (
+  <button>
+    <PlusIcon id="unique-plus-icon" /> New
+  </button>
+)
+```
```
joshblack, line 127: Similar to the `ref` one above, I bet with `id` we could just treat it as typical DOM props that consumers would expect to be able to provide and understand what it can be used for. Curious what your thoughts are here 👀

joshblack (commented):
Looking great 🥳 When you get a sec, could you merge in `main`? Would love to see what it looks like with some of the `ref` changes that came up in latest. Left a couple comments and questions too, let me know what you think!

lib/octicons_react/README.md
green6erry: (We just removed the example all together)

lib/octicons_react/README.md
```diff
@@ -121,6 +158,13 @@ export default () => (
 )
 ```
 
+<<<<<<< HEAD
+
+> ⚠️ The `Octicon` component is deprecated. Use icon components on their own instead:
+
+=======
+
+> > > > > > > origin/main
```
joshblack, line 167:
```suggestion
```

joshblack (approved):
Sorry for the delay! Just left one comment for the README, seems like there are still diff markers there but otherwise should be good to go!

PR #943, opened 2023-05-03
Update Icons to use `React.forwardRef` & Remove support for `Octicon`

### LINKED ISSUE #910: Update React components to use `React.forwardRef()`
Currently our Icon components are exported as plain functional components. It would be great if we included support for forwarding refs through `React.forwardRef()`, however this is only available past [React v16.3](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#forwardref-api). Currently, the project lists React v15 as a peer dependency.

It seems like we have a couple of paths forward:

- Increase the peer dependency range, but still include v15, and try and feature detect `forwardRef`. This would land in a `minor` change to the package
- Increase the peer dependency range explicitly to >=v16.3 and unconditionally use `forwardRef`. This would land in a `major` change to the package

broccolinisoup:
Closes #910

As stated in the issue itself, our icon components do not forward their refs and this is currently causing issues in the remediated Tooltip implementation https://github.com/primer/react/pull/3032. It is also a great enhancement for future cases to make sure their refs are available to access when needed.

Re introducing this change as a major, please see [the comment](https://github.com/primer/octicons/issues/910#issuecomment-1511659451)

### DIFF
--- a/lib/octicons_react/README.md
+++ b/lib/octicons_react/README.md
@@ -121,9 +121,6 @@ export default () => (
 )
 ```
 
-### `Octicon` (REMOVED)
-
-> ⚠️ The `Octicon` component is removed. Use icon components on their own instead:
 
 ```diff
 - <Octicon icon={AlertIcon} />
--- a/lib/octicons_react/README.md
+++ b/lib/octicons_react/README.md
@@ -121,9 +121,10 @@ export default () => (
 )
 ```
 
-### `Octicon` (DEPRECATED)
+### `Octicon` (REMOVED)
 
-> ⚠️ The `Octicon` component is deprecated. Use icon components on their own instead:
+> ⚠️ The `Octicon` component is removed. Use icon components on their own instead:
+
 ```diff
 - <Octicon icon={AlertIcon} />
 + <AlertIcon />
--- a/lib/octicons_react/src/index.js
+++ b/lib/octicons_react/src/index.js
@@ -1,12 +1,1 @@
-import React from 'react'
-
-export default function Octicon({icon: Icon, children, ...props}) {
-  // eslint-disable-next-line no-console
-  console.warn(
-    // eslint-disable-next-line github/unescaped-html-literal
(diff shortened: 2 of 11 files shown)


lib/octicons_react/src/index.js
```diff
@@ -6,7 +6,7 @@ export default function Octicon({icon: Icon, children, ...props}) {
     // eslint-disable-next-line github/unescaped-html-literal
     '<Octicon> is deprecated. Use icon components on their own instead (e.g. <Octicon icon={AlertIcon} /> → <AlertIcon />)'
   )
-  return typeof Icon === 'function' ? <Icon {...props} /> : React.cloneElement(React.Children.only(children), props)
+  return typeof Icon === 'object' ? <Icon {...props} /> : React.cloneElement(React.Children.only(children), props)
```
broccolinisoup, line 9: I am a little unsure about this change 😵‍💫 Would there be any impact on consumer applications? 🤔

broccolinisoup: Thanks for your great suggestion @joshblack! I removed the deprecated `Octicons` 🎉  Build and tests seems fine but let me know if there is anything I am missing to do here.

Also, I pushed it as a commit here, would that be better if I pushed another PR for it you think? 🤔

broccolinisoup: If it is good to go within this PR, I'll update the title and the changeset. If not, I'll push another PR

lib/octicons_react/README.md
```diff
@@ -121,9 +121,10 @@ export default () => (
 )
 ```
 
-### `Octicon` (DEPRECATED)
+### `Octicon` (REMOVED)
+
+> ⚠️ The `Octicon` component is removed. Use icon components on their own instead:
```
joshblack, line 126: I'd just remove this section if the package no longer offers this as an API.

```suggestion
```

joshblack (approved):
Just left one comment for the README! Otherwise, looks great!

colebemis (approved):
🚀

Issue #940, opened 2023-04-10
[Icon request] Unpin


yp05327:
### Describe the topic

Could you add an unpin icon?

### Anything else?

Thank you for this pretty icon set

silverwind:
Current pin icon is

<img width="238" alt="Screenshot 2023-04-17 at 17 08 01" src="https://user-images.githubusercontent.com/115237/232527911-127be359-93c2-463f-a99c-4845c57b7542.png">

My idea would a same icon, but with a strike-through, similar to [these](https://www.svgrepo.com/vectors/unpin/monocolor/):

<img width="317" alt="Screenshot 2023-04-17 at 17 08 27" src="https://user-images.githubusercontent.com/115237/232528242-3cab5433-3086-4f29-935d-db9771fa327b.png">

CameronFoxly:
I've started thinking about adding this request to the library, and realized that we'll have to flip the direction of our `pin` icon, so that the slash follows the same directional conventions (slashing from top left to bottom right) of the rest of the library. This also actually seems like the right move for the `pin` icon because more of the icons in the library that have a diagonal composition tilt the way the the flipped version of the `pin` icon would (with the tip of the pin pointing left instead of right). 

Here's my proposed solution for both an updated `pin` and the new `unpin`, at 2x and 1x in both 16px and 24px sizes:

![Image](https://user-images.githubusercontent.com/19292210/235740578-738bc3fb-9ec1-4296-8007-8b5a54843f61.JPG)

@tallys, can you check with product owners to make sure swapping the direction of the `pin` icon would be problematic?

tallys:
Will follow up! Thanks @CameronFoxly

CameronFoxly:
A new [release](https://github.com/primer/octicons/releases/tag/v19.1.0) went out this morning which has included the `pin-slash` icon. It can be seen [here](https://primer.style/design/foundations/icons/pin-slash-16) and is now included in the library and npm package. Thanks for the request, @yp05327. 

Closing this out 🚀

PR #950, opened 2023-05-10
Save pin-16.svg, pin-slash-16.svg, pin-24.svg, pin-slash-24.svg

CameronFoxly:
This PR adds `pin-slash` (alt keyword `unpin`) and flips the orientation of our current `pin` icon to make them work together. Discussion in [this issue.](https://github.com/primer/octicons/issues/940)

Screen Shot 2023-05-10 at 11 34 40 AM


colebemis:
@CameronFoxly I think you need to remove the frame background in Figma. When the SVG optimizer runs, the frame background turns into a black square: 
CleanShot 2023-05-10 at 11 57 33@2x

CameronFoxly:
Thanks for the headsup @colebemis. Should be good now!

PR #966, opened 2023-07-26
Fix `pin-slash` bug

gavinmn:
Looks like there was a bug in these Octicons where the `16px` version was `17px` wide and the `24px` version was `25px` wide. This PR fixes that.


CameronFoxly (approved):
LGTM, thanks for taking this on!

PR #971, opened 2023-08-07
Use the Primer GitHub App for auth instead of the GPR_AUTH_TOKEN_SHARED

camertron:
This PR replaces usage of the `GPR_AUTH_TOKEN_SHARED` PAT with a token obtained through the Primer GitHub App in several CI scenarios.

Fixes https://github.com/github/primer/issues/2540

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,6 +33,15 @@ jobs:
           yarn
           bundle install
 
+      - id: get-access-token
+        uses: camertron/github-app-installation-auth-action@v1
+        with:
+          app-id: ${{ vars.PRIMER_APP_ID_SHARED }}
+          private-key: ${{ secrets.PRIMER_APP_PRIVATE_KEY_SHARED }}
+          client-id: ${{ vars.PRIMER_APP_CLIENT_ID_SHARED }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -33,6 +33,15 @@ jobs:
           yarn
           bundle install
 
+      - id: get-access-token
+        uses: camertron/github-app-installation-auth-action@v1
```
rezrah, line 37: 👋 @camertron - what do you think about pulling this workflow into our [Primer org](https://github.com/primer/.github/tree/main/.github/workflows)? I think it could help centralise our workflows and make it a little easier to maintain.

camertron: Yeah that's a great idea! Let me try that.

camertron: Hey @rezrah, I finally got around to trying this out and ran into a problem. It appears Actions does not allow sharing outputs between workflows from different repos if those outputs contain secrets. Actions automatically detected the access token secret I was trying to send over and skipped exporting it to the calling action. I tried to use the private key to encrypt the token in the shared action and decrypt it in the calling action. The whole point of this exercise was to make the config easier to maintain though, and the encrypt/decrypt approach is arguably much more complex than what we already have.

PR #977, opened 2023-08-17
Add new Feed icons

gavinmn:
Closes https://github.com/github/primer/issues/2529

Adds a handful of new Feed icons

CleanShot 2023-08-17 at 10 21 14@2x


marcalcobe:
Hi @gavinmn and @tallys ,

When adding this new icons to our library of Octicons in Figma (to keep it updated), I've noticed that two of the issue icons are called the same "feed-issue-open" resulting with the package only containing the icon that looks like reopen.

What is the normal procedure in this cases? Should I open a separate issue for this?

gavinmn:
@marcalcobe thank you for reporting this! I should have a fix ready to go in https://github.com/primer/octicons/pull/980

PR #980, opened 2023-08-22
Fix Feed icon alignment issue and name collision

gavinmn:
Fixes the naming issue on `feed-issue-reopen` and `feed-issue-open` caused by them both being named `feed-issue-open` before. 

Context https://github.com/primer/octicons/pull/977#issuecomment-1687846472

Also including an alignment fix for the other Feed icons. I think the Octicons plugin upload `17px × 16px` icons again. This should move all to be `16px × 16px`.


icons/feed-issue-reopen-16.svg
CameronFoxly, line 1: Looks like this one is another 17 x 16px version 🤦 

I wonder why that keeps happening...

CameronFoxly (approved):
LGTM! 🚢

PR #990, opened 2023-10-13
add azure icon to Octicons

pansyng:
@arisacoba has created Azure Icon for vnet public beta use. Proposing to add to Octicons for future usage. 
Following contributing guideline:
- added azure icon in icons folder 
- update keyword.json


arisacoba:
Wanna clarify that I didn't create this but [it's from Azure](https://www.iconcloud.design/search/azure/063030d91-de64dd68d) 😅

gavinmn:
@tallys what's the latest on including branding elements like this in Octicons?

tallys:
hi @pansyng we do not include nor open source logomarks or brand elements in Octicons - the infrastructure for that is Octovisuals and we are still setting that up. Our recommendation is to use the icon in place as is, and we can include this icon in our plans for Octovisuals as we get that up and running.

Issue #699, opened 2021-11-20
[Feedback] npm package contains duplicate icon data


xt0rted:
### Describe the topic

I just installed v16.1.1 of [@primer/octicons](https://www.npmjs.com/package/@primer/octicons) and noticed the `svg` folder and `data.json` file are duplicated in the package. I don't think this is intended and it's doubling the size of the package.

<img width="167" alt="image" src="https://user-images.githubusercontent.com/831974/142723128-d5176714-9a76-4470-a7c4-f34d1c78f4e9.png">

ashygee:
Hi @xt0rted, thanks for letting us know! We'll take a look at this once we get some open bandwidth.

xt0rted:
Still an issue in 17.2.0

lesliecdubs:
@jonrohan is this something you might be able to squeeze in this quarter?

tallys:
Following up, @jonrohan is this still an issue or can we close this?

jonrohan:
I think this is still an issue. I believe both data sources are in use, so we'd need to investigate which one we can remove and update any libraries using them.

tallys:
Hey @JoshBowdenConcepts, this might be a good one to pick up - thoughts?

tallys:
Thank you for raising this issue, it is a known error and is in our backlog.

fregante:
Would you accept a PR? Sounds like a quick change. This bug doubles the size of the package

https://packagephobia.com/result?p=@primer/octicons@13.0.0

lesliecdubs:
Hi @fregante, thanks for the offer. Yes, we would gladly accept a PR for this change!

fregante:
@lesliecdubs done

lesliecdubs:
Thanks @fregante! ✨ We might be a little slower than usual with PR review due to the holiday season, but we should be able to get to it in the new year. Thanks for contibuting!

fregante:
This can be closed once https://github.com/primer/octicons/pull/1010 is published to npm https://github.com/primer/octicons/commits/main/

joshblack:
Just wanted to confirm that I don't see this anymore in v19.9.0. Going to close this out as I believe it's been addressed, feel free to leave a comment if I'm getting something wrong here!

fregante:
Indeed in my link you can see the package going from 2.16 to 1.08, exactly half

PR #1001, opened 2023-12-20
Avoid duplicate build folder in @primer/octicons

fregante:
- Probably fixes https://github.com/primer/octicons/issues/699 

In my local tests, running `yarn build` in `lib/octicons_node` generates `build/build/svg` but after this change it generates `build/svg`, so it should fix the issue.

However the build is a bit of spaghetti between workflows, docker entrypoints.js and multiple package.json, so it's not entirely clear if further work is required (probably not)

### DIFF
--- a/lib/octicons_node/package.json
+++ b/lib/octicons_node/package.json
@@ -17,7 +17,7 @@
     "url": "https://github.com/primer/octicons/issues"
   },
   "scripts": {
-    "build": "\\cp -r ../build/ ./build && \\cp index.scss ./build/build.css",
+    "build": "\\cp -r ../build/. ./build && \\cp index.scss ./build/build.css",
     "lint": "eslint index.js tests/*.js",
     "test": "ava --verbose 'tests/*.js'"
   },


lib/octicons_node/package.json
```diff
@@ -17,7 +17,7 @@
     "url": "https://github.com/primer/octicons/issues"
   },
   "scripts": {
-    "build": "\\cp -r ../build/ ./build && \\cp index.scss ./build/build.css",
+    "build": "\\cp -r ../build/. ./build && \\cp index.scss ./build/build.css",
```
fregante, line 20: See https://stackoverflow.com/a/4645159/288906

fregante:
@broccolinisoup ☺️

camertron (approved):
Nice, thank you!

camertron:
I'm not sure what's going on with these build failures, but I assume it's because this PR was created from a fork.

camertron:
Merged in #1001

PR #1007, opened 2024-03-04
fix(octicons_react): remove role if aria-hidden

joshblack:
This is a small changes that makes `role` conditional on if `aria-label` or `aria-labelledby` is present. This is to avoid having both `aria-hidden` and `role` on the element at the same time.

This also updates the root `package.json` file so that contributors with yarn >2 don't mess with the lockfile locally


broccolinisoup (approved):
Looks great 🎉

keithamus (approved):

PR #1009, opened 2024-03-04
fix(octicons_react): remove build from npm package files

joshblack:
Closes https://github.com/primer/octicons/pull/1002

This removes `build` from the list of `files` for the npm package. These files are not accessible since we currently use the `exports` field so removing this should be compatible within semver (but let me know if not!)

### DIFF
--- a/lib/octicons_react/package.json
+++ b/lib/octicons_react/package.json
@@ -25,7 +25,6 @@
     "posttest": "yarn ts-test"
   },
   "files": [
-    "build",
     "dist"
   ],
   "keywords": [


broccolinisoup (approved):
>These files are not accessible since we currently use the exports field so removing this should be compatible within semver (but let me know if not!)

That makes sense to me!

fregante:
Hmm why not just merge my PRs?

camertron:
@fregante I'm not 100% sure in this case, but we usually do this because certain actions won't run from forks.

joshblack:
Hi @fregante, I'm so sorry about that, I wasn't aware of it. Totally my fault!

fregante:
No problem to leave as is, I was just wondering

PR #1012, opened 2024-03-13
Remove Octicon example from react README.md

smithamax:
Looks like 697bf97f1dcd6490dff1515e976b311460680b15 only deleted the section title, not the actual section in the README


keithamus (approved):
Thanks for this!

PR #1015, opened 2024-03-14
Fix versioning script that updates the version files

jonrohan:
There's a lint error in #1014 that is caused because when changesets is versioning it's removing the lint comment https://github.com/primer/octicons/commit/9f09f452d8376158380938e84d2691e5f084a175


joshblack (approved):

camertron:
Thanks for this! 🎉

PR #1020, opened 2024-04-15
fix(octicons_react): include extension in type declaration

### LINKED ISSUE #1018: [Bug] Esm requires file extension in d.ts files.
Typescript projects using the moduleResolution `NodeNext` or `node16` cannot consume `@primer/octicons-react` 19.9.0 (and transitively `@primer/react`36.13.0, regressing from 36.12.0) since the icons type definition is not specifying a file extension.

The import will just fail and, when peeking into the index.d.mts, you'll find the error:
```
relative import paths need explicit file extensions in ecmascript imports
```

Patch that worked for me:

```diff
diff --git a/dist/index.d.mts b/dist/index.d.mts
index acfe71a1ce71b4764c3d155334b3bbc2830ea5d0..8539c08534ff1732a8f80c3ba850dfd6cc7f4ac8 100644
--- a/dist/index.d.mts
+++ b/dist/index.d.mts
@@ -2,7 +2,7 @@
 import * as React from 'react'
 
 // eslint-disable-next-line prettier/prettier
-import { Icon } from './icons'
+import { Icon } from './icons.js'
 
 type Size = 'small' | 'medium' | 'large'
 
@@ -20,4 +20,4 @@ export interface OcticonProps {
   verticalAlign?: 'middle' | 'text-bottom' | 'text-top' | 'top' | 'unset'
 }
 
-export * from './icons'
+export * from './icons.js'
```

joshblack:
Closes #1018 

Add an explicit extension for the types to support different `moduleResolution` options. This also adds in a step to the CI to run `arethetypeswrong` to make sure that we don't regress here in the future 👀

### DIFF
--- a/.changeset/blue-crabs-tie.md
+++ b/.changeset/blue-crabs-tie.md
@@ -1,0 +1,5 @@
+---
+'@primer/octicons': patch
+---
+
+Update types for @primer/octicons-react to explicitly include extensions for different moduleResolution settings
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,11 +90,8 @@ jobs:
         run: yarn test
       - name: Types
         run: |
-          yarn pack
-          VERSION=$(jq -r '.version' package.json)
-
           # Run @arethetypeswrong/cli
-          npx @arethetypeswrong/cli --pack 'primer-octicons-react-v$VERSION.tgz'
+          npx @arethetypeswrong/cli --pack .
 
           # Run publint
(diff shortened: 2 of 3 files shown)


siddharthkp (approved):
Oooh nice!

Just double checking, is this safe to publish as a patch release?

joshblack:
@siddharthkp I believe so since this will be fixing the issue to what the intent was (right now it's broken 😅). Let me know if there is something in particular that we should be guarding against / looking for 👀

PR #1023, opened 2024-05-02
Rename some wrongly named icons

lukasoppermann:
We have other icons like `circle` and `circle-slash`. In Figma these icons are also called `circle-check`. It feels wrong to invert the naming logic for this one icon.

The alternative would be to go the other way and rename the remaining icons and all figma icons to `[thing]-circle`.

> [!Warning]
> **Warning:** This is probably a breaking change. We could mitigate this by:
> 1. Releasing the renamed and old icons in a minor update.
> 2. Remove the badly named icons in a major release
> 3. Search and replace all instances we find on github and primer.


offensive-vk (approved):

Haxxor247 (approved):

mperrotti (approved):

lukasoppermann:
@camertron, pinging you as I hope you can help.

We are currently in a tough spot, were renaming icons is nearly impossible, as we need to update the entire consuming project at the same time.

I think we need an approach that allows us to deprecate icons but keep them in package somehow.

Basically, having two versions of an icon, but one comes from a deprecated folder? 

I don't know if we can log a console message for deprecated icons, e.g. `The icon "iconName-16" is deprecated, please use "newIconName-16" instead.

This would allow us to have both versions in the package at the same time, so that we can update the consumers and remove the icons in a next major. Any thoughts?

camertron:
@lukasoppermann hmm interesting...

> Basically, having two versions of an icon, but one comes from a deprecated folder?

That should be possible, yes 👍 

> I don't know if we can log a console message for deprecated icons, e.g. `The icon "iconName-16" is deprecated, please use "newIconName-16" instead.

I'm not sure about React, but this is definitely possible for the Rails version of the octicons gem 😄

lesliecdubs:
@lukasoppermann do we know how critical this change is for consumers? How often do we hear confusion/trouble due to the wrong naming scheme? 

I know we value consistency, but given that we'd need to scope and prioritize this against other pressing needs I'm interested to understand the demand for and impact of this change.

cc'ing @sunnyi101 as well, since it sounds like this might come down to a prioritization decision

PR #1039, opened 2024-11-11
Update file-media and image icons

nataliakeniganti:
The current file-media and image octicons have always felt a little too cartoony to me and don't match the style of the rest of the icons. I took a stab at redesigning them:


colebemis:
Looks like the only change in this PR is extra XML tags in the SVG file.

PR #1041, opened 2024-11-29
Add new sparkles icon

lukasoppermann:
This icon was request by @dthoma1.

Fixes https://github.com/github/primer/issues/4430


colebemis (commented):
Should we call this `sparkles-fill`? That leaves the option to have an outline version in the future and is consistent with the existing `sparkle-fill` icon: https://primer.style/foundations/icons/sparkle-fill-16

dthoma1:
This icon fits our needs for the docs ai assistant project. ✨

lukasoppermann:
@colebemis is there a naming convention for icons somewhere?

tallys:
@colebemis you are a gem and an eternal champion of Octicons, thank you for reviewing!
@lukasoppermann thanks so much for pushing this through to support @dthoma1 !!!

tallys (approved):

PR #1063, opened 2025-01-31
Add square-circle icons

dylanatsmith:
This PR adds `square-circle-16` and `square-circle-24` to be used in github.com surface areas inside IconButton as stop buttons. Consensus is that the existing `square-fill` can be too ambiguous.

Naming `square-circle` was chosen to be consistent with `x-circle` and `plus-circle`.

Note: @CameronFoxly chose the 0.75px corner radius for the 16px icon.

[Source file](https://www.figma.com/design/2rjsQp42uKvzPXzLZIqOeA/Octicons-v2---Source?node-id=9827-3)


CameronFoxly (approved):
LGTM! Approved once we fix those build fail issues

dylanatsmith:
Opened https://github.com/primer/octicons/pull/1064 to fix the CI failure

francinelucca (approved):

PR #1105, opened 2025-07-23
Bump node-version to 22.x for GitHub workflows

myabc:
- **Bump Bump ava to 6.4.1**
- **Update test snapshots**
- **Bump node-version to 22.x for GitHub workflows**


myabc:
Please let me know if you need a changeset.

PR #1165, opened 2026-02-28
Addition of lockup icon, adjustments to logo and mark

kylewaynebenson:
## Mark 24
| Before | After |
|--------|--------|
| image | image |

## Mark 16
Before was slightly stretched to pixel fit 16x16, after restores to un-stretched circle

| Before | After |
|--------|--------|
| image | image |

## Logo 24

Icon is slightly wider to keep height match with old logotype.

| Before | After |
|--------|--------|
| image | image |

## Logo 16

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/.changeset/olive-ties-crash.md
+++ b/.changeset/olive-ties-crash.md
@@ -1,5 +1,0 @@
----
-'@primer/octicons': patch
----
-
-various dep updates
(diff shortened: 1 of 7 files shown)


simurai (approved):
Ok, I changed the version to `minor` since we add an additional icon.

Otherwise, I think this is good to ship? 🚢 

/cc @lukasoppermann @francinelucca

PR #1193, opened 2026-04-22
Stack icon variants

kylewaynebenson:
Expanding on the stack icon to include a remove and check variant to support Stacked PRs in symbolizing unstacking.
## stack-remove
image

## stack-check
image


siddharthkp (dismissed):
Hi! Something went wrong here, there are 700 files changed!

Maybe more files got generated than we need? Looking at previous PRs adding icons, that wasn't the case.
- https://github.com/primer/octicons/pull/1165
- https://github.com/primer/octicons/pull/1157

kylewaynebenson:
@siddharthkp It was the [optimize SVGs](https://github.com/primer/octicons/blob/main/.github/workflows/optimize.yml) action that added/re-optimized the other 698 icons, but it looks like that action hasn't been updated in 4 years—so I don't totally understand why it's suddenly misfiring.

TylerJDev (approved):

PR #1263, opened 2026-07-27
Add default exports to per-icon octicons-react modules

mattcosta7:
Add default exports for per-icon modules in @primer/octicons-react while keeping root/barrel APIs named-only.


francinelucca (commented):
confused about what's happening here 👀 . what's this PR doing?

francinelucca (approved):

PR #1316, opened 2026-07-31
Update execa

joshblack:
This PR updates `execa` and migrates the build tests to its current API. Part of dependency stack #1323.

### Changed

- Update `execa` from 4.0.0 to 10.0.1.
- Load the ESM-only package with a dynamic import.
- Update failure assertions for the current `ExecaError` shape.

This is an internal test tooling change and does not change the public API.

### DIFF
--- a/tests/build.js
+++ b/tests/build.js
@@ -1,5 +1,11 @@
 const path = require('path')
-const execa = require('execa')
+
+let execa
+
+beforeAll(async () => {
+  const execaModule = await import('execa')
+  execa = execaModule.execa
+})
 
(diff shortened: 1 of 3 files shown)


tests/build.js
```diff
@@ -1,5 +1,11 @@
 const path = require('path')
-const execa = require('execa')
+
+let execa
+
+beforeAll(async () => {
+  const execaModule = await import('execa')
+  execa = execaModule.execa
+})
```
joshblack:
```suggestion
const { execa } = await import('execa')
```

PR #1203, opened 2026-05-11
Add flag octicon

ericwbailey:
This PR adds `flag-16` and `flag-24` icons to Octicons, following [this request](https://github.com/github/primer/issues/6624).

The current working draft is contained in [this Figma branch](https://www.figma.com/design/1ljgTFkT5NKNRfq5hw07JQ/branch/mhRdvKjVP8Nr9KHjS53gD5/Octicons?m=auto&node-id=176-0&t=VhhYO0rGcLH0w3DB-1), which will be merged when this PR is merged.

PR #1329, opened 2026-08-04
Add issue-relates-to octicon

skw:
This PR adds `issue-relates-to-16` and `issue-relates-to-24` to Octicons.

The icon is a counterpart to [`issue-tracked-by`](https://primer.style/octicons/icon/issue-tracked-by-16/) with the arrow flipped to point outward, indicating that an issue links *out* to something else. It is intended for use in the "Relates to" issue relationship filter suggestions in github/issues.

- Original icon drawn by @eggimo
- Requested for github/issues#23343

### Preview

_(Screenshot predates the rename from `issue-relates` to `issue-relates-to`; the artwork is unchanged.)_

issue-relates-to

cc @eggimo @khiga8


khiga8 (commented):
Looks good to me! should we  name this `issue-relates-to`?

.changeset/clever-donuts-relate.md
```diff
@@ -0,0 +1,7 @@
+---
+'@primer/octicons': minor
+'@primer/octicons-react': minor
+'@primer/styled-octicons': minor
```
skw: I added these to the changeset per copilot's feedback, but should double check that this makes sense as I'm not super familiar with how these changesets work.

skw: I dug into this and reverted it back to only `@primer/octicons` in 808d1a2d — I don't think Copilot's suggestion applies to how this repo actually versions things.

Versions for the React/styled packages aren't derived from changesets at all:

- [`script/version`](https://github.com/primer/octicons/blob/main/script/version) (run by the `changeset:version` npm script) force-syncs the root, node, **and react** `package.json` versions from `@primer/octicons`.
- [`publish.yml`](https://github.com/primer/octicons/blob/main/.github/workflows/publish.yml) publishes all three npm packages at the **root** `package.json` version — including `@primer/styled-octicons`, whose own `16.1.0` is overwritten by `npm version --allow-same-version "$PACKAGE_VERSION"`.

So naming them doesn't change what gets released. It only adds CHANGELOG entries, under version headers computed independently by changesets that don't match the published versions — `lib/octicons_react/CHANGELOG.md` already tops out at `## 19.31.1` while its `package.json` is `19.32.0`. And `@primer/styled-octicons` has no `CHANGELOG.md` at all, so listing it would create one.

That matches precedent too. Across every changeset in the repo's history: `@primer/octicons` named 35 times, `@primer/octicons-react` 4 times (all React-runtime changes — aria-hidden, ESM, subpath exports — never icon additions), `@primer/styled-octicons` 0 times. The most recent icon addition, #1203, used only `@primer/octicons`.

Happy to be overruled if the intent is for the CHANGELOGs to list icon additions per-package, but that looks like it'd be a broader change.

skw:
@khiga8 good call — renamed to `issue-relates-to` in 808d1a2d.

joshblack (approved):

